Issue #5, opened 2014-08-26
Add OSGi manifest headers


mattbishop:
I'd like to use Byte Buddy in an OSGi application but cannot because the jar does not contain the necessary META-INF/manifest headers.

The easy way to add them is to use the maven-bundle-plugin.

ui4j:
Hi,

We are unable to use bytebuddy for OSGi environments. Are you sure that Export-Package definition is correct?

```
Export-Package: !net.bytebuddy.jar.asm.*,net.bytebuddy.*
```

Related issue and comment:
https://github.com/ui4j/ui4j/issues/28#issuecomment-106741600

raphw:
Not entirely sure. To be honest, I do not use OSGi, I only googled my way. I am going to have a look at it.

CodingFabian:
I guys, i will fix this, but I want to get the requirements straigt. 
Right now only byte-buddy-agent produces valid osgi headers, however I assume this is not even included as bundle, because its a javaagent. This is not even a valid bundle (because it doesnt provide services) so I propose to remove it.
byte-buddy-dep and byte-buddy both do not produce osgi headers.

Could you please clarify which Classes you actually want to use form which jar?

CodingFabian:
Update, i am building the 0.7 snapshot and the osgi headers are valid. i am able to load the bundle in karaf:
karaf@root()> bundle:install mvn:net.bytebuddy/byte-buddy/0.7-SNAPSHOT
Bundle ID: 53
53 | Active    |  80 | 0.7-SNAPSHOT   | Byte Buddy (without dependencies)
karaf@root()> bundle:info 53

Byte Buddy (without dependencies) (53)

raphw:
The current version or your build?

CodingFabian:
without my changes. It works, but as sent to you via skype, a really correct header is pretty long, due to the amount of packages and tangling you have there

CodingFabian:
Here are the new manifest files my upcoming PR will produce. Would these work for you?
https://gist.github.com/CodingFabian/35f727056830156b2756

raphw:
Version 0.6.7 should have correct OSGi headers. Thanks @CodingFabian for fixing this!

raphw:
On second look, it seems like the shaded POM is not longer overriding the original POM. The dependency remains declared...

raphw:
[...]

PR #29, opened 2015-05-29
Create valid OSGI Headers for byte buddy.

CodingFabian:
This change introduces the maven-bundle-plugin to generate the correct
MANIFEST.MF values. Due to the internal structure of byte buddy, these
are pretty long. The build procedure also generates a few warnings about
nonstandard osgi behaviour, which is ok, but might encourage future
refactorings.

### DIFF
--- a/byte-buddy-agent/pom.xml
+++ b/byte-buddy-agent/pom.xml
@@ -9,11 +9,10 @@
     </parent>
 
     <artifactId>byte-buddy-agent</artifactId>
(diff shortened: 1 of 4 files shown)


CodingFabian:
this should fix #5,
the manifests generated are here: https://gist.github.com/CodingFabian/35f727056830156b2756

@raphw I checked the jars, and they look ok, maybe you want to doublecheck that inclusions into the final jars did not change.

PR #34, opened 2015-07-06
FixedValue.value(String.class) returns Class<Class> instead of Class<String>

gsson:
When passing a class to FixedValue.value() load the passed class instead of the class of the passed class.


gsson:
Workaround for anyone not using master:

    FixedValue.value(new TypeDescription.ForLoadedType(theClass));

PR #50, opened 2015-08-02
Improve Readme Example (fixes #49)

### LINKED ISSUE #49: Obscure documentation
Hi Rafael,
i just noticed that you use quite an obscure example in the 3-1 = 2 comparison example in the main readme.
It is correct, but it evokes a wtf moment when you see that it substracts hash codes. Is that intended, or shall I send a PR to make that less magical?

CodingFabian:
Note: my editor (atom) removed trailing spaces, if you do not want to have this in the commit, let me know and I fix this up. You can compare the PR ignoring whitespace here: https://github.com/raphw/byte-buddy/pull/50/files?w=1

I made the example easier to understand in various aspects.
First I called the method `compare`, rather than `intercept`. In fact there are existing references to `ComparisonInterceptor::compare` in the Readme already, so I assume that was a non synchronized change.
Second, I thought it should strictly follow the contract of compare by returning -1, 0, or 1, as well as being less magical about hashCodes. So I just delegate to the compareTo method of comparable. This could throw class cast, but I still think it is easier to understand (and class cast is even documented in Comparator.compare())
I also used numbers which are not so confusingly close to each other. (so you wont confuse it with a subtract by accident)

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ runtime code generation for the Java virtual machine
 [![Build Status](https://travis-ci.org/raphw/byte-buddy.svg?branch=master)](https://travis-ci.org/raphw/byte-buddy)
 [![Coverage Status](http://img.shields.io/coveralls/raphw/byte-buddy/master.svg)](https://coveralls.io/r/raphw/byte-buddy?branch=master)
 [![Maven Central](https://maven-badges.herokuapp.com/maven-central/net.bytebuddy/byte-buddy-parent/badge.svg)](https://maven-badges.herokuapp.com/maven-central/net.bytebuddy/byte-buddy-parent)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -93,36 +93,37 @@ signature of the `Comparator:compare` method we want to implement later:
...
-passing an instance of the above class to `MethodDelegation`: 
+that by Byte Buddy generates require Byte Buddy on the class path! We can now implement the `Comparator` interface by
+passing an instance of the above class to `MethodDelegation`:
 
 ```java
 Class<? extends Comparator> dynamicType = new ByteBuddy()
-  .subclass(Comparator.class) 
-  .method(named("compare")).intercept(MethodDelegation.to(new ComparisonInterceptor())
+  .subclass(Comparator.class)
+  .method(named("compare")).intercept(MethodDelegation.to(new ComparisonInterceptor()))
```
CodingFabian: note that the missing closing brace should be fixed even if this pr is not accepted :-)

raphw:
This is way better. Thanks for doing this!

PR #71, opened 2015-10-14
Added isAbstract and isInterface convenience ElementMatchers

CodingFabian:


CodingFabian:
was created mostly using copy and paste, let me know if I missed a place where i should have copied and pasted as well :)

raphw:
Thanks for that! Could you fix the trailing spaces and ammend the commit to your push? You can run alls tests locally with `-Pchecks -Pintegration`.

CodingFabian:
didn't know about -Pchecks, thanks. Eclipse is pretty stubborn, do you happen to have an Eclipse formatter? Or only idea? :)

raphw:
I use IntelliJ. The `checks` profile should be active by default.

PR #72, opened 2015-10-14
Do not retransform an empty array.

CodingFabian:
Transforming an empty array/list would throw:
*** java.lang.instrument ASSERTION FAILED ***: "numClasses != 0" at JPLISAgent.c line: 1102


CodingFabian:
If you suggest a different code style, I can do that of course.

raphw:
I though the agent API was smarter. This is exactly what I would do, thanks for that, all is fine. Please fix the one test case and ammend the change.

CodingFabian:
Yeah sorry, didn't check for tests. You mocked the agent API incorrectly, lets see if i can get mockito to throw a segfault :)

raphw:
No need for that. Simply change the Mockito assertion for the invocation of the method which is still required for the empty array.

PR #75, opened 2015-10-19
Fix net.bytebuddy.implementation.attribute.AnnotationAppender.ValueFilter.SkipDefaults#sRelevant(...)

rgoldberg:
`SkipDefaults#isRelevant(...)` should include non-default values, thus it should return `true` iff the value is not a default.

The existing logic was backwards.

### DIFF
--- a/byte-buddy-dep/src/main/java/net/bytebuddy/implementation/attribute/AnnotationAppender.java
+++ b/byte-buddy-dep/src/main/java/net/bytebuddy/implementation/attribute/AnnotationAppender.java
@@ -343,7 +343,7 @@ public interface AnnotationAppender {
             @Override
             public boolean isRelevant(AnnotationDescription annotationDescription, MethodDescription.InDefinedShape methodDescription) {
                 Object defaultValue = methodDescription.getDefaultValue();
-                return defaultValue != null && defaultValue.equals(annotationDescription.getValue(methodDescription));
+                return defaultValue == null || !defaultValue.equals(annotationDescription.getValue(methodDescription));
             }
 
             @Override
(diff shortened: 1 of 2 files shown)


raphw:
Good catch, thanks. Could you add a unit test that verifies this logic? Also, please remove the white space that checkstyle complains about and ammend the commits. (Simply push force to your branch and your pull request is automatically updated.

rgoldberg:
I don't know why the build is failing.  It looks like the `org.apache.maven.plugins:maven-checkstyle-plugin:2.15` plugin is causing the failure, but the output isn't very helpful.

raphw:
It says:

```
/home/travis/build/raphw/byte-buddy/byte-buddy-dep/src/main/java/net/bytebuddy/implementation/attribute/AnnotationAppender.java:346:49: '!' is followed by whitespace.
```

rgoldberg:
Just saw your comments.  I'll redo my whitespace.  checkstyle really should output that as an error, rather than an info.  It was buried among tons of other info logs...

raphw:
Thank you. Could you please squash the commits and force push this to your branch? This would allow me to easier merge this to the current development branch.

rgoldberg:
Squashed.

When do you think 0.7 will be officially released (non RC)?

raphw:
Thanks.

Difficult to say. There is still quite a lot to do, unfortunately. I have to rework the internal representation of the type system what will require a refactorization of a large amount of code. Also, I am currently not having too much free time to work on it. I hope to releasing it early next year. Generics has been an insane amount of work, but I want to do it properly in order to allow for the integration of value types and primitive generic types etc. in the future which will be a more pressing use-case.

PR #90, opened 2016-01-13
Add onArgument target for method invocations

antigremlin:
A parameter index must be specified as .onArgument(0) and the
invocation target will be extracted from the intercepted method
parameters. A new TargetHandler is introduced to produce the
necessary bytecode.


raphw:
The change is now part of 1.0.0 - for your case, you do however need to enable dynamic assignment. Thanks for the suggestion! Cheers.

PR #122, opened 2016-04-30
Don't use TypeDescription#getInternalName but #getName in equals to improve performance a bit

felixbarny:
All getInternalName does is calling `getName().replace('.', '/')` anyway

This was revealed by using Flight recorder.

getinternalname


raphw:
Thanks, good catch. I did not realize that I did not do this change already. Long time ago, I streamlined everything to rely on binary names instead of internal names as a loaded intances cannot provide an internal name without transformation.

PR #154, opened 2016-07-10
Added additional checking for common errors during compile time using google errorprone

CodingFabian:
Google Errorprone (http://errorprone.info/) is a small enhancement of the javac compiler to catch more programming mistakes. While some could be considered also "style" (like lowercase L for marking longs) it can never harm to have an additional level of checking.


raphw:
Unfortunately, error prone only works with JDKs 8+ while Byte Buddy triggers CI builds on all VMs of version 6 and older. This would require a seperate compilation profile and I think any actual bug should be covered by a test.

CodingFabian:
:-( Yeah, thought you use 8 to compile backwards to 6. Was not aware the build runs on 6

raphw:
Google has no culture for supporting old versions of anything. ;)

PR #215, opened 2016-11-19
Revert TerminationHandler.resolve() back to public.

peihe:
We hit this backward incompatible change when we upgrade. It blocks us from upgrading to mockito 2.

Was reducing visibility intentional, and what is the alternative?

It is used in:
https://github.com/apache/incubator-beam/blob/master/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/reflect/ByteBuddyDoFnInvokerFactory.java#L469

It surprised me that backward incompatible changes are made within one major version. Followings two are also backward incompatible but easier to fix:
MethodVariableAccess.loadOffset() -> loadFrom()
MethodReturn.returning() -> of()

### DIFF
--- a/byte-buddy-dep/src/main/java/net/bytebuddy/implementation/bind/annotation/TargetMethodAnnotationDrivenBinder.java
+++ b/byte-buddy-dep/src/main/java/net/bytebuddy/implementation/bind/annotation/TargetMethodAnnotationDrivenBinder.java
@@ -546,7 +546,7 @@ public class TargetMethodAnnotationDrivenBinder implements MethodDelegationBinde
          */
         RETURNING {
             @Override
-            protected StackManipulation resolve(Assigner assigner, MethodDescription source, MethodDescription target) {
+            public StackManipulation resolve(Assigner assigner, MethodDescription source, MethodDescription target) {
                 return new StackManipulation.Compound(assigner.assign(target.isConstructor()
                                 ? target.getDeclaringType().asGenericType()
                                 : target.getReturnType(),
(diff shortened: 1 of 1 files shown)


raphw:
First of all, I apologize for these changes. I sometimes forget that more people then me are using the byte code level API. The particular renamings I made in an airplane when I got annoyed that statements did not fit on my screen so I removed some naming redundancies.  Of course, this is not a good enough reason to change namings and I will remember this in the future.

For some of the primitives, the changes were however necessary as I refactored Byte Buddy to work with generic types where the least-specific representation has become a `TypeDefinition` rather than a `TypeDescription`. I did not catch all the places where I should have refactored what is mostly a performance improvements, I just did not have the capacities.

Until now, I felt like nobody but me used these low-level representations but I guess the library has grown in a way where I should have realized that I cannot change a few things and then simply update the usage in the code bases of all my users. Again, I am sorry for the inconvenience. I just went through the APIs and improved a few last changes; I promise to attempt better compatibility in the future.

As for the use of the `TerminationHandler`, this change was intended. The termination handler was added to support the fluent API of a method delegation. It does not make sense to use it outside of the context of such a delegation as the implementation might change with the method delegations generated byte code. You should not rely on this. I would therefore replace it with the method body:

```java
new StackManipulation.Compound(assigner.assign(target.isConstructor()
    ? target.getDeclaringType().asGenericType()
    : target.getReturnType(),
  source.getReturnType(),
  RuntimeType.Verifier.check(target)), MethodReturn.of(source.getReturnType()));
```

If you can guarantee that the target is no constructor or/and if you are not using the `@RuntimeType` annotation, you can simplify this to:

```java
new StackManipulation.Compound(
  assigner.assign(target.getReturnType(), source.getReturnType(), Assigner.Typing.STATIC),
  MethodReturn.of(source.getReturnType())
);
```

Not reading the annotation will probably improve performance quite a bit.

peihe:
Thanks Rafael for the detailed response and suggestions.
That should work for us.

(closing this PR)

PR #231, opened 2016-12-12
Added support for loading arguments from an array.

cemelo:
This is probably not the best way to achieve this, since it has an specific requirement: the typing must be dynamic.


raphw:
Thanks, I will have a look as soon as I find some time!

raphw:
Thanks for your PR. I refactored it a bit and changed some of the naming. Appreciate your effort and will release this soon.

cemelo:
Thank you for this awesome library.

PR #242, opened 2016-12-29
Upgrade ASM to 5.2

Godin:
ASM 5.2 contains some nice fixes - see http://mail.ow2.org/wws/arc/asm/2016-12/msg00013.html


CodingFabian:
Changes from 5.1:

- Better comments in Opcodes.java
- SuppressWarnings in Opcodes.java for Integer constructor deprecation.
- More tests to check binary backward compatibility
- Bug fixes:
   #317616 Misleading comment in ClassWriter
   #317621 Type.getReturnType(desc) returns wrong for types with
parenthesis in name
   #317749 ClassCheckAdapter should discover invocation of default
method from pre-Java 8 class file
   #317626 AnnotationNode incorrectly process array values in
'visit(final String name, final Object value)' method
   #317630 Frames are re-calculated for Large Methods
   #317748 Class after instrumenting loses first BootstrapMethods entry

PR #248, opened 2017-01-06
Issue 245

orange-buffalo:
Introduced Lombok, replaced `equals` and `hashCode` method with `@EqualsAndHashCode` annotation.
Did not apply refactoring for cases where:
* (abstract) getters are used
* methods are marked with `@SuppressFBWarnings` (as I did not have time to check thoroughly how FindBugs is invoked and if the change will have affect on reporting)
* custom logic is applied
* "outer" object is used
* Lombok is confused (rzwitserloot/lombok#1274)  
* tests expect current implementation

Removed `toString()` methods but the places where tests expect specific implementation.

Btw, I have notices that class `ObjectPropertyAssertion` is duplicated in several modules. Maybe it makes sense to create a separate module with common test code (including this class) and use this module as a test dependency in all other places?


raphw:
Awesome, thank you. I will try to review this sometime next week. Quite a lot of code changes but I trust its good for! Excited to see the impact on the code size.

Cheers, Rafael

PR #261, opened 2017-02-08
Prevent redefining 0 types in resubmission

CodingFabian:
This will prevent the message "Redefined 0 types in 0 batches"


raphw:
This should be taken care of by the collector. Did you encounter an error?

PR #265, opened 2017-02-19
Fix broken build plugins with failing tests

mches:
This fixes java.lang.IllegalArgumentException: foo/Bar contains the illegal character '/' (on Windows). 'file' is a relativized URI#toString(), not File#toString(), therefore the separator character will be '/' on any platform (including Windows).


raphw:
Good catch, thanks!

mches:
No prob. Thank you for Byte Buddy.

PR #273, opened 2017-03-07
Add AppVeyor (Windows-based) CI

mches:
This should reduce the likelihood that Windows-specific bugs are released. You would just have to enable [AppVeyor CI](https://ci.appveyor.com/login).


raphw:
Awesome, thank you! I develop on Linux, this was on my mind before and I will merge this asap.

Out of curiosity: why is the Maven wrapper necessary? Does AppVeyor not supply a default installation?

mches:
So, AppVeyor CI provides Maven 3.3.9 currently, which is not Java 6-compatible. Travis CI provides 3.2.5 currently, but no doubt they will eventually upgrade to 3.3.  There are some methods of scripting that could install Maven, but it varies by CI platform. Maven Wrapper seems to be really popular. It's based on Gradle Wrapper which is practically a de facto standard. Jason van Zyl is key maintainer, so that helps its credibility quite a bit. Travis CI actually detects Maven Wrapper and uses it if you override the default scripts. Anyone who clones a repository with Maven Wrapper can run with the correct Maven version. Just replace `mvn` with `./mvnw`.

raphw:
I see, thanks for the explanation and your effort!

mches:
After you enable AppVeyor CI, you'll have something like this: 

image

PR #332, opened 2017-08-15
Removed check that local variables in frame have the same types as parameters

elizarov:
This check is too strict and was rejecting valid class files.
It fixes mockito issue 1152: https://github.com/mockito/mockito/issues/1152


raphw:
As explained in the Mocktio issue, this cannot be done.

PR #343, opened 2017-09-11
Use wildcard in Advice.WithCustomMapping#bind

felixbarny:
I think `OffsetMapping.Factory<? extends Annotation>` would be redundant as the generic parameter T can only extend Annotation because of the interface declaration `interface Factory<T extends Annotation>`. Also, the generic type parameter is not important in the method. That's why I went for `OffsetMapping.Factory<?>`.

### DIFF
--- a/byte-buddy-dep/src/main/java/net/bytebuddy/asm/Advice.java
+++ b/byte-buddy-dep/src/main/java/net/bytebuddy/asm/Advice.java
@@ -8553,7 +8553,6 @@ public class Advice implements AsmVisitorWrapper.ForDeclaredMethods.MethodVisito
          * a parameter of an advice method, the dynamic value is asked to provide a value that is then assigned to the parameter in question.
          *
          * @param offsetMapping The dynamic value that is computed for binding the parameter to a value.
-         * @param <T>           The annotation type.
          * @return A new builder for an advice that considers the supplied annotation type during binding.
          */
         public WithCustomMapping bind(OffsetMapping.Factory<?> offsetMapping) {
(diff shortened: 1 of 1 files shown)


felixbarny:
Strange errors during build:

```
net.bytebuddy.build.gradle.ByteBuddyExtensionTest > testTaskExplicitExcluded STANDARD_ERROR
    Sep 11, 2017 9:39:41 AM org.codehaus.groovy.runtime.m12n.MetaInfExtensionModule newModule
    WARNING: Module [groovy-all] - Unable to load extension class [org.codehaus.groovy.runtime.NioGroovyMethods]
Gradle Test Executor 1 finished executing tests.
[INFO] Execute :test
*** buffer overflow detected ***: /usr/lib/jvm/java-6-openjdk-amd64/bin/java terminated
======= Backtrace: =========
/lib/x86_64-linux-gnu/libc.so.6(__fortify_fail+0x37)[0x7fd9e208fdb7]
/lib/x86_64-linux-gnu/libc.so.6(+0x10bcb0)[0x7fd9e208ecb0]
/usr/lib/jvm/java-6-openjdk-amd64/jre/lib/amd64/libnet.so(Java_java_net_Inet4AddressImpl_getLocalHostName+0x1a0)[0x7fd9d15f9df0]
[0x7fd9d87d86b0]
```

Have you seen that before?

raphw:
I had a lot of trouble with Travis lately. It is mostly due to their distribution change for the build environment and I assume this is releated. I will have a look!

felixbarny:
I know that feel. Having the same trouble with travis atm...

PR #406, opened 2018-02-09
Fix typos in comments

hunterwb:
Fixes typos in Javadoc and inline comments


raphw:
Wow, thanks for doing that!

PR #409, opened 2018-02-15
Detect maven compiler target java version in multi module projects

DiabolusExMachina:
This will detect the java target version in multi module maven projects. 
Will break ByteBuddyMojoTest.java


raphw:
Looks good but could you avoid the whitespace changes?

DiabolusExMachina:
I think the format should look ok now.

raphw:
I cleaned up a bit and merged your PR.

PR #435, opened 2018-04-07
Reduce memory allocation for ArrayProjection.getName()

testn:
ArrayProjection.getName() allocates a few extra objects. 
This change eliminates most of them by pre-allocate
the exact StringBuilder size and perform the replacement
manually instead.


raphw:
Hey, thanks for your PR. Did you measure that this has an actual impact? There are plenty minimal improvements that I skipped to keep the code base cleaned up. This is not to bad and maybe worth it due to the high reliance on names throughout the library. Was that your thought?

raphw:
Merged manually.

PascalSchumacher:

PR #436, opened 2018-04-07
Reduce memory allocation in CompoundList

testn:
CompoundList is often called with empty list.
This change will reduce the memory allocation 
in that case.


raphw:
If the list is empty, it would probably be most efficient to return a singleton list returning the single element?

testn:
Yes, it is. If possible, I actually want to refactor the entire library. It overuses ArrayList everywhere. If we make the library to use something similar to ImmutableList, it will make the optimization much easier.

raphw:
Merged manually.

PascalSchumacher:

PR #437, opened 2018-04-07
use for instead of foreach to reduce extra memory allocations

testn:


testn:
@raphw  you have a lot of wrapper class around `List<XXX>` which I finds it pretty useful to convert one type to another. However, when you use a wrapper around list, it seems to disable the optimizer to optimize away the iterator. In addition to that, sublist() seems to create ArrayList$SubList and wrapped behind another wrapper. This actually causes some performance problem when having to load a bunch of methods.

PascalSchumacher:

raphw:
I always tag the last release version as I sometimes need to break API and then would release 1.9.0.

PR #438, opened 2018-04-07
Add TypeDescription.ForLoadedType.of which will return an instance of

testn:
Add TypeDescription.ForLoadedType.of which will return an instance of well known type when possible to reduce memory allocation. We can convert most of the call to use this method instead of the constructor later.

### DIFF
--- a/byte-buddy-dep/src/main/java/net/bytebuddy/description/type/TypeDescription.java
+++ b/byte-buddy-dep/src/main/java/net/bytebuddy/description/type/TypeDescription.java
@@ -68,6 +68,86 @@ public interface TypeDescription extends TypeDefinition, ByteCodeElement, TypeVa
     TypeDescription VOID = new ForLoadedType(void.class);
 
     /**
+     * A representation of the {@code Integer} non-type.
+     */
+    TypeDescription INTEGER = new ForLoadedType(Integer.class);
+
+    /**
+     * A representation of the {@code Long} non-type.
+     */
+    TypeDescription LONG = new ForLoadedType(Long.class);
+
+    /**
+     * A representation of the {@code Short} non-type.
+     */
+    TypeDescription SHORT = new ForLoadedType(Short.class);
+
+    /**
+     * A representation of the {@code Byte} non-type.
+     */
+    TypeDescription BYTE = new ForLoadedType(Byte.class);
+
+    /**
+     * A representation of the {@code Character} non-type.
+     */
+    TypeDescription CHARACTER = new ForLoadedType(Character.class);
+
+    /**
+     * A representation of the {@code Double} non-type.
+     */
+    TypeDescription DOUBLE = new ForLoadedType(Double.class);
+
+    /**
+     * A representation of the {@code Float} non-type.
+     */
+    TypeDescription FLOAT = new ForLoadedType(Float.class);
+
+    /**
+     * A representation of the {@code Boolean} non-type.
+     */
+    TypeDescription BOOLEAN = new ForLoadedType(Boolean.class);
+
+    /**
+     * A representation of the {@code int} non-type.
+     */
+    TypeDescription INTEGER_PRIMITIVE = new ForLoadedType(int.class);
+
+    /**
+     * A representation of the {@code long} non-type.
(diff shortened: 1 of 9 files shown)


testn:
failed due to a findbug issue https://github.com/spotbugs/spotbugs/issues/500

byte-buddy-dep/src/main/java/net/bytebuddy/description/type/TypeDescription.java
```diff
@@ -7079,6 +7184,14 @@ public static String getName(Class<?> type) {
                     : name.substring(0, anonymousLoaderIndex);
         }
 
+        public static ForLoadedType of(Class<?> type) {
```
raphw: This should probably just return a TypeDescription. The implementation should not be exposed.

byte-buddy-dep/src/main/java/net/bytebuddy/description/type/TypeDescription.java
```diff
@@ -67,6 +67,111 @@
...
+     */
+    TypeDescription FLOAT_PRIMITIVE = new ForLoadedType(float.class);
+
+    /**
+     * A representation of the {@code boolean} non-type.
+     */
+    TypeDescription BOOLEAN_PRIMITIVE = new ForLoadedType(boolean.class);
+
+    Map<java.lang.reflect.Type, ForLoadedType> WELLKNOWN_TYPE_MAP =
+        new HashMap<java.lang.reflect.Type, ForLoadedType>() {
```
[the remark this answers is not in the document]
raphw: Should probably also go into the `ForLoadedType` class.

byte-buddy-dep/src/test/java/net/bytebuddy/description/type/TypeDescriptionForLoadedTypeOfTest.java
```diff
@@ -0,0 +1,151 @@
...
+import org.junit.Ignore;
+import org.junit.Test;
+
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+public class TypeDescriptionForLoadedTypeOfTest extends AbstractTypeDescriptionTest {
```
raphw: A particular test is not necessary here, this is an implementation detail that should be able to be switched out without noticable changes.

testn:
I need to move everything into ForLoadedType as the inner class of an interface is loaded before the interface itself. After that, I alias everything back to TypeDefinition.XXXX to avoid backward compatibility problem

raphw:
@testn Out of curiosity: Are you applying these changes based on profiled runs? Did this improve performance for your AssertJ performance issue? Just checking, these changes are not intrusive but decrease the code quality a tiny bit every time, therefore I only want to add them if they have proven useful.

testn:
Introduce `TypeCache` nested class to break out of a recursive class loading problem.

testn:
@raphw I just wonder how I debug bytebuddy via intellij. It looks like it keeps trying to compile bytebuddy code without running it through bytebuddy-maven-plugin

raphw:
You can bind the maven goal for the Byte Buddy plugin to the make command. Then it is always executed. It should however not be necessary for most cases, the essential hashCode/equals methods must be implemented explicitly.

testn:
@raphw  have you got a chance to take a look? I didn't replace all the `new ForLoadedType` usage. Just the necessary one.

raphw:
I am currently busy with a sick kid but I will look at it.

As said, I was wondering if you and how you did benchmark this. Just checking to make sure that I do not complicate the code base for no reason,

raphw:
@testn Just to make sure if you read this section. How did you benchmark this? I am wondering since this avoids additional allocations but adds an additional lookup for all classes that are not in the cache so I wonder if this really pays off.

testn:
With this change, `SoftAssertionsPerfTest` takes 4.5 seconds down from around 6 seconds. It seems to save over 200,000 allocations on that.

PR #439, opened 2018-04-07
Use much more efficient method to determine whether the class is a package-info class

testn:
This is to optimize validated() method which will 
call into this method very often. As package-info string must 
be the end of full class name, we can optimize away the need to 
call String.indexOf/String.substring


raphw:
I will not merge this either since validation is a developer configuration that should not be used in production and to keep the code simple.

PR #441, opened 2018-04-08
Improve validated() call by skipping HashSet creation when possible

testn:


raphw:
Since validation is a debugging configuration, I will not merge this to keep the code simpler.

PR #446, opened 2018-04-16
Do not leak classes via NexusAccessor

CodingFabian:
before this change there was an exception attached to the unavailable NexusAccessor.

This exception is either a problem, or it was constructed when injection was disabled.

Unfortunately, this stack trace captures all the classes that were on the stack when the NexusAccesor was first time used.
Since this is just for information only, this PR changes it to a String reason, instead of the stack, thus removing the leakage of classes.


CodingFabian:
screen shot 2018-04-16 at 12 14 16
illustration of the problem found in a heap dump

PR #455, opened 2018-04-25
Optimize special case of OfTypeVariable.ForLoadedType.equals

testn:
It is called a lot by findBindingOf when implementing generic types


raphw:
Thanks, I will implement such shortcuts for all loaded type representations to keep it consistent. Thank you!

PR #481, opened 2018-05-30
Replace 'install-jdk.sh' by new jdk constants

sormuras:
Travis CI now uses 'install-jdk.sh' internally.

See https://github.com/travis-ci/travis-build/pull/1347


raphw:
That is great, makes things much easier! Well done!

PR #502, opened 2018-07-21
Apache SkyWalking use case

wu-sheng:
Add SkyWalking java agent to use case. @raphw Thank you so much to maintain this cool project.


raphw:
I am glad you like it! The list in the readme is not meant to be inclusive, it just mentions a few early adopters. I am happy to list Skywalking in the Wiki!

wu-sheng:
Sure. Thanks anyway.

PR #506, opened 2018-08-03
Let ElementMatchers::isClone match when overridden with different return type.

bwaldvogel:


raphw:
You are right, this check slipped in by accident.

Issue #321, opened 2017-07-06
Allow MethodCall instance to be chained on their result


raphw:
Add an API to `MethodCall` to allow chaining invokations on the previous result.

raphw:
Implemented on master.

Issue #275, opened 2017-03-10
#question#How to invoke method ON the return value of another method call?


jujusharp:
Now, I wanna to generate a class like:

     public Object someMethod(){
           return getSomeProperty().callOtherMethod();
     }

I can generate the code call with "field" in the class with:

     MethodCall.invoke(otherMethod).onField("someField")

and here is the code generated:

     public Object someMethod(){
           return someField.callOtherMethod();
     }

I found `WithoutSpecifiedTarget`.on() support call with **class instance**, **argument** and **field**, but can't support `method return value`. Is there any way to do this?

raphw:
Unfortunately, there is currently no good way of doing so. You can generate such more complex code by defining a custom `ByteCodeAppender`. I consider adding something viable in the future.

jujusharp:
Got it, Thanks!

liuchang0403:
i think it is not nesseray. byte buddy is a easy lib compare to ASM, i think it is enough now.

raphw:
Captured by #321

PR #522, opened 2018-08-25
Implement MethodCall chaining

mibac138:
Added `MethodCall.onMethodCall` and `MethodCall.withMethodCall` methods along with tests and documentation for both of them. 
As I have found no other workaround, I added `Target`  as a parameter to both `ArgumentLoader.Factory.make` and `TargetHandler.resolve` because generating the `StackManipulation` for `MethodCall` requires it.

A possible further enhancement is supporting the following syntax: `MethodCall.invoke(named("foo")).onField/Arg/MethodCall(...)` so that the function `foo` is searched inside the type passed through the `onX` method but it requires additional work. If I were to do it I'd like your input as to whether it should be added.

Related issue: #321, #275


raphw:
Nicely done, thanks for that!

PR #523, opened 2018-08-27
Avoid leaking matchers via resubmission

CodingFabian:
screen shot 2018-08-27 at 10 09 40
This screenshot from Eclipse MAT shows that its possible to leak Matcher instances (and with it a whole tree of Instrumentation things) when the Resubmission Executor is reused after an instrumentation has `reset`.

The reason is that its possible that the `ResubmissionInstallationListener` (implements Runnable) still sits in the Executor queue.

To avoid that, we should also release all references to heavy classes (like Matchers) after cancel is invoked.

Please rename/reformat to your liking.

### DIFF
--- a/byte-buddy-dep/src/main/java/net/bytebuddy/agent/builder/AgentBuilder.java
+++ b/byte-buddy-dep/src/main/java/net/bytebuddy/agent/builder/AgentBuilder.java
@@ -5803,7 +5803,7 @@ public interface AgentBuilder {
                 /**
                  * A job that resubmits any matched type that previously failed during transformation.
                  */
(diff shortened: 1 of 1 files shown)


CodingFabian:
This has a very similar result as `java.util.concurrent.ScheduledThreadPoolExecutor.setRemoveOnCancelPolicy(boolean)`
which is unfortunately a Java 7 API

raphw:
I think the worst leak is that of the class loader which cannot be prevented this way. A best solution removes the future from the root loader which can be done by implementing a custom dispatching scheduler.

Issue #515, opened 2018-08-10
Garbage collection pressure created by Method.getParameters()


dsyer:
I see that `Method.getParameters()` is implemented internally in the JDK as a `clone()` and this generates a lot of garbage collection pressure with Hibernate because of the way that Byte Buddy iterates over method parameters (I think), and also because Hibernate adds static methods to entity proxies with large numbers of parameters.

The offending code path is 

```
ParameterList$ForLoadedExecutable$OfMethod.get(int) line: 320	
ParameterList$ForLoadedExecutable$OfMethod.get(int) line: 307	
AbstractList$Itr.next() line: 358	
TargetMethodAnnotationDrivenBinder.compile(MethodDescription) line: 64	
MethodDelegation$ImplementationDelegate$ForStaticMethod.of(MethodList<?>, MethodDelegationBinder) line: 759	
MethodDelegation$WithCustomProperties.to(TypeDescription) line: 1223	
MethodDelegation$WithCustomProperties.to(Class<?>) line: 1204	
MethodDelegation.to(Class<?>) line: 255	
```

the problem being that iterating over all method parameters causes a new instance of `ParameterDescription.ForLoadedParameter.OfMethod` to be created for each parameter, and each one takes a copy of _all_ the parameter types via the JDK `Method` implementation. I imagine the iteration could be optimized.

I am using Flight Recorder to look at garbage collection pressure which slows down application startup for large (in this case pathological) apps - i.e. with large numbers of entities. I think if it could be optimized here it would have a beneficial effect on less pathological apps.

raphw:
Thanks for reporting. `MethodDelegation` instances are quite expensive by themselves as they process metadata which is another weak spot in the JVM performance-wise. I will have a look how this is used in Hibernate, the instances are immutable and can be reused.

dsyer:
You are welcome.

Another source of GC pressure is the internal implementation of `ParameterDescription.ForLoadedParameter.OfMethod`. For instance `getDeclaredAnnotations()` re-parses the annotations of the whole method for every parameter, I think. And `getType()` clones the parameter types again. The former is definitely visible in the Flight Recorder as a creator of `char[]` in the `ConstantPool`, while the latter is not yet visible, but might surface if we are optimizing other things. I can open a separate issue about that if you prefer.

gsmet:
@dsyer do you have the rest of the stacktrace, the Hibernate part?

gsmet:
[...]

PR #529, opened 2018-08-28
Avoid calling Method.getParameters more than one time per method

mibac138:
Added caching of parameter annotations.

Related issue #515

### DIFF
--- a/byte-buddy-agent/pom.xml
+++ b/byte-buddy-agent/pom.xml
@@ -5,7 +5,7 @@
     <parent>
         <artifactId>byte-buddy-parent</artifactId>
         <groupId>net.bytebuddy</groupId>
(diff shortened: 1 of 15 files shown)


raphw:
The problem with this approach is that it reads the annotations eagerly while this is not required most of the time, I want to avoid that.

mibac138:
I've now changed it so that annotations are read lazily.

raphw:
This approach would not be thread-safe. I am aiming in a different direction here. I just added a new plugin for caching method return values and I will add something here in the line of what I started to add to the project. 

I think I want to cache annotations on the basis of the `MethodDescription` instance that emits the `ParameterDescription`. In addition, I will try to make better use of the `Parameter` API where annotations can be read more selectively.

PR #564, opened 2018-11-14
Cosmetic improvement of the Matchers look during debugging: anyOf and noneOf

Sanne:
When debugging non trivial Matcher trees being represented as string, this helps being a little less verbose.

Should have a positive impact on performance too, although likely too small to be measureable.


Sanne:
thanks!

PR #644, opened 2019-04-29
Configure Maven build scans

aalmiray:
Build scans (http://scans.gradle.com/) capture build data when a build is run, then you can inspect what happened. You can think of build scans are x-rays on a build. Here's a sample report generated with the changes from this branch https://scans.gradle.com/s/gcuo24gxfjtvm

Please refer to the Terms of Service of this free service at https://gradle.com/terms-of-service
If needed, the agreement of terms can be made conditional using environment variables.


raphw:
Thanks for that. I will need to look into this in a bit more detail and see if I really need this extension before I merge it. The Byte Buddy build is already fairly complex, I'd rather reduce the tooling required to run it at this point, especially considering the new release cycle of Java where I basically need to update everything all of the time.

aalmiray:
Great. Let me know if there's additional information needed and I'll provide it. The maven extension for build scans collects information without interfering with the build, this means that if the extension were to fail it does not make the build fail.

aalmiray:
Closing due to inactivity. Feel free to reach out to Gradle's Dev Advocacy should you like to continue.

raphw:
Sorry, I have not had time lately. It is on my list, I will look at it eventually. Thanks!

PR #653, opened 2019-05-14
Add OpenJ9 11 Travis build

szpak:
Based on some closed issues it seems that you try to support OpenJ9. Therefore, it would be probably good to have that build on CI server. I added it on Travis.
There is at least one failing [test](https://travis-ci.org/szpak/byte-buddy/jobs/532165974#L3849-L3867). Maybe it would be necessary to add an extra profile to cover some other differences.

As an additional negative effect the Travis build will be slightly slower in general as only 5 builds can be executed in parallel in the open source plan.


raphw:
This is great, thanks!

PR #655, opened 2019-05-14
Mark OpenJ9 Travis build as "allow_failures"

szpak:
Just in a case you want to have green Travis builds until fixed/enhanced :).

image


raphw:
Not needed, just one test that is now fixed.

PR #668, opened 2019-06-12
Adjust gradle plugin readme

epeee:
a minor one, it should rather be `Gradle Plugin` instead of `Maven Plugin`.


raphw:
Thanks, good catch!

PR #704, opened 2019-08-05
Check JNA version when resolving emulated VM

felixbarny:
The emulated attach requires JNA 5 but Byte Buddy will happily resolve the emulated `VirtualMachine` and fail with a `NoSuchMethodError` if JNA 4 is present. This can happen quite unexpectedly when inheriting from Spring Boot's parent pom, as it overrides the JNA version. See also https://github.com/elastic/apm-agent-java/pull/775


raphw:
The implementation works with JNA 4 and 5 quite easily if one just avoids using the introduced `load` method. I rather avoid using the method and use the deprecated `loadNative`.

felixbarny:
I tried that but had problems with the apis used for windows like CreateSemaphore. I saw you solved that already. Great job, thanks ❤️

PR #754, opened 2019-10-26
Fixed ASM guide URL

ineuwirth:
Fixed ASM guide URL, the previous one hosted on http://forge.objectweb.org is no longer available, see: https://www.ow2.org/view/IT_Infrastructure/GForge_discontinued


raphw:
Thanks, very appreciated.

PR #755, opened 2019-10-26
Fix exception message in AgentBuilder.RedefinitionStrategy.RETRANSFORMATION

ineuwirth:
(redefinition -> retransformation)


ineuwirth:
The Javadoc of `RETRANSFORMATION` might need to be revised too, the `Note` section seemed to be a simple copy from `REDEFINITION`.

PR #788, opened 2019-12-10
Update Gradle build

melix:
This commit reworks how the Gradle plugin is built. Instead of using
different versions of Gradle for the purpose of testing (which doesn't
really work because the plugin doesn't do functional testing but uses
the API jar), the same version of Gradle is used (the latest, 6.0.1).

However, this requires a bit of tweaking:

- Gradle will always be executed with a Java 8 JVM, which requires
to set the `JDK_8` environement variable
- when Maven calls Gradle, it tells what is the target bytecode
version, which is then used by Gradle to determine what JDK to use.
It will, for example, use JDK 7 to compile and execute the tests,
using forking. Again, an environment variable `JDK_7` needs to be
set.

The difficulty of this procedure is that this build is building a
Gradle plugin. As such, it relies on the `gradleApi` jar, which is
a generated jar at runtime. This jar contains more than the API,
as the name doesn't imply, because it also contains the minimal
runtime for testing with TestKit.

However, this plugin does _not_ use TestKit. It just runs tests
and checks against the Gradle API. Ideally, this Gradle API should
be published and reusable, and this plugin should be able to be
built for older Gradle versions. This is what the build tries to
emulate by tweaking the Gradle API jar in order to lower its
bytecode version.

This, however, raises different questions:

- if the plugin is built with Java 8, it is likely to depend on
internal details of the API, making it incompatible with previous
Gradle releases, so what's the point of testing?
- if the plugin is built with a previous format, like it does in
this build, then we can execute it on older Gradle releases (likely),
but then we don't test that it will _also_ work with newer JDKs.

For this, Gradle would make it much easier by allowing you to create
different test suites for the different target JDKs. The fact that the
build is driven by Maven doesn't make it easy, but, let's be honest,
the fact that Gradle doesn't publish API jars is also painful.

It would be worth investigating _functional_ testing instead of unit
testing to test the behavior against different Gradle versions, and
in particular look at the https://gitlab.com/ysb33rOrg/gradleTest
plugin.


raphw:
Thanks a lot again. I finally decided against the class file version back-patching. I tried it in Byte Buddy and it is a pain and I cannot guarantee that it works reliably in the future.

In the end, I want to be sure that the plugin is build-able on any JVM version. It might sound silly but it made the build quite robust in the past and I want to keep it that way without excluding old versions where a lot of people really rely on Byte Buddy, also compared to other tools. In the end, I cannot guarantee that Java 8 is installed in all test environments which are often isolated in containers.

I took a lot of your suggestion and hope to have fixed the dependency issue. Thanks a lot for your effort, I learned a bunch about Gradle, too!

melix:
Np. A few comments. I still think using 2 versions of Gradle is not a good idea. I agree it's currently way too complicated to support multiple versions of Gradle for a plugin, but it's not uncommon. Data we get is that the vast majority of users are on Gradle 4+, which means Java 8.

Second, I think you could backport the `buildSrc` trick to make your build cleaner (the repetition of the whole JavaVersion block isn't nice).

I still think you're doing it wrong in terms of testing. What you're doing is _not_ unit testing, it's functional testing. You want to check that ByteBuddy behaves correctly under Gradle version X, for JDK Z. The appropriate tool for this is the Tooling API, which is capable of downloading Gradle releases for you and execute arbitrary builds on it.

You could also leverage [this plugin](https://gitlab.com/ysb33rOrg/gradleTest) to make cross-version testing easier.

Eventually I still don't understand why you want to initialize ByteBuddy _within the Gradle process itself_. As far as I understand the goal of this plugin is to transform classes produced by the build, but not the build itself. You have to remember that Gradle uses the Gradle Daemon, which is a long living process aimed at executing builds. If you transform classes within this process, then _all subsequent builds_ will see the transformations, which breaks the principle of isolation. It's also possible that the daemon is reused for other builds that do _not_ use this plugin.

I feel like what you really want is a task which transforms the classes at build time (compile time to be precise), which can be achieved without leaking anything into Gradle, for example by using the [worker API](https://docs.gradle.org/current/userguide/custom_tasks.html#worker_api). Gradle provides a wide variety of options here, including classloader and process isolation modes. It's worth noting that the worker API supports running on older JDKs (down to 6 if I recall).

raphw:
There is a chance that I am doing it wrong. To be honest, I just googled "Gradle plugin" and tried to make something similar to what I already did for Maven. You are saying the approach is wrong? In a way, the plugin takes one or multiple "user plugins" that perform Byte Buddy-based transformations onto the build's classes or test classes after regular compilation. Could you point me to what other API I would be supposed to implement here? Is it this worker API? So far no user reported any problems.

And yes, the testing could be improved, I just did not find the time yet. As for the usage statistics. I am fully aware of that. As for people who invest money in open source, it is overly people who have some wish to guarantee support for legacy systems, I try to also cater to those to increase the chances of the project's survival.

melix:
Yes, I'm saying that the user provided transforms should be executed in a sandbox after compilation. Those transforms shouldn't really depend on Gradle APIs. This sandboxing mechanism is exactly what the worker API provides: the transform would be executed in isolation in a separate process, where you can use whatever JDK you want, independently of the JDK Gradle in executed with. See https://guides.gradle.org/using-the-worker-api for examples.

Note, however, that the worker API only supports JDK 6+.

PR #799, opened 2019-12-28
Patch api

raphw:


felixbarny:
Hi Rafael 🙂 

Some questions: 
- What are some of the intended use cases for this?
- What is the advantage of patching an existing `ResettableClassFileTransformer` vs creating another one?
- Can a patched `ResettableClassFileTransformer` be patched again?
- It seems like the type matchers for the existing instrumentations are executed again, is that intended? Seems like it can be quite expensive.
- Instead of creating a new AgentBuilder, would it be possible to add additional transformations to an existing `ResettableClassFileTransformer`?
- Should you create a new AgentBuilder with only the transformers you want to add or augment an existing AgentBuilder when patching a `ResettableClassFileTransformer`?
  - The former does not seem to work when patching a `ResettableClassFileTransformer` multiple times - the transformations of the former patches seem to be lost.
  - When doing the latter, the classes are matched on each patch


What I'm trying to do is to ensure a class has been instrumented at runtime and re-transform it if it's not instrumented yet. The idea behind is to get rid of expensive matchers like `hasSuperType(named("org.asynchttpclient.AsyncHandler"))` in favor of ensuring `AsyncHandlers` to be instrumented which are provided to the `org.asynchttpclient.DefaultAsyncHttpClient#executeRequest(Request request, AsyncHandler<T> handler)` method.

The reason matchers like `hasSuperType(named("org.asynchttpclient.AsyncHandler"))` are so expensive is that we can't really pre-filter which classes have to be scanned. This means that we have to scan the full inheritance hierarchy of all loaded classes which is a bit like a shotgun-surgery. Matching for `org.asynchttpclient.DefaultAsyncHttpClient#executeRequest(Request request, AsyncHandler<T> handler)` is more like a laser surgery as we can exactly match the method by class name and method name.

See also https://github.com/elastic/apm-agent-java/compare/master...felixbarny:patch-agent?expand=1

Is `patchOnByteBuddyAgent` even the right tool for that?

raphw:
So the idea is about as follows. You have an agent builder that creates a class file transformer that would match the following classes to those transformers:

> Foo.class -> transformer1
> Bar.class -> transformer2
> Qux.class -> transformer1, transformer2
> Baz.class -> { }

Now you create a new class file transformer with another agent builder that matches the following:

> Foo.class -> transformer1
> Bar.class -> transformer1,transformer2
> Qux.class -> {}
> Baz.class -> transformer3

You want to replace the previous class file transformer with the latter one; this can be the result of for example changing a configuration file or similar. Previously, it was required to reset the first class file transformer before installing the second one which would require six retransformations in two batches. With patching, you can reduce the amount of retransformations to three in a single batch and not even touch *Foo.class* as its state has not changed.

You can patch a class file transformer as many times as you want. This would work as follows:

```java
class PatchableUnit {
  ResettableClassFileTransformer transformer;

  void install(AgentBuilder builder, Instrumentation inst) {
    if (transformer == null) transformer = builder.installOn(inst);
    else transformer = builder.patchOn(inst, transformer);
  }
  void undo(Instrumentation inst) {
    if (transformer != null) {
      transformer.reset(inst, RedefinitionStrategy.RETRANSFORMATION);
      transformer = null;
    }
  }
}
```

All you do is to recreate your class file transformer from scratch as you would normally do where you include all retransformations. If you want to add additional instrumentations to those you already installed, it is equally cheap to install an additional transformer with a new agent builder that focuses on these additional transformers as long as you make sure those instances share caches and so forth. The idea of patching is to replace an existing agent builder.

I'd say the typical use case is where you add multiple transformations to an agent builder but subsequently add new transformers or refine previous ones. Byte Buddy now allows you to refine your code and to patch your existing transformation without computing the diff manually. If you could already do this easily, e.g. if you only ever add new transformations steps without replacing or altering previous ones, simply install additional transformers which should be the cheapest way of doing so.

To compute this diff, Byte Buddy needs to reapply all matchers; it is however efficient about it when your ignore matchers are good enough and those matchers are also reapplied on loaded types where hierarchy matches are normally rather simple to do. The application is also conducted in the attach thread (or any thread spawned by it) what should not distort the host process and its user threads. Those are normally quite impacted by additional retransformations which require a stop-the-world pause while being applied which should be reducable by a lot when patching small diffs.

I understand that you'd like to get rid of hierarchy-parsing matchers but it is almost impossible to do so. If `Foo extends Bar`, then `Foo` is loaded before `Bar`. So, at the time you needs to make a decision, it is not yet possible to decide based on a loaded constant. You could technically throw an exception on every non-loaded type and use resubmission to only match types in hierarchies but due to the reverted loading of type trees, I'd say chances are good to end up in a dead-lock at some point.

felixbarny:
Thanks, that was really insightful!

So it seems it is highly recommended to use `AgentBuilder.DescriptionStrategy.Default.HYBRID` and `POOL_ONLY` would not be a good fit as it would load all the TypeDescriptions from the file system again, is that right?

> I understand that you'd like to get rid of hierarchy-parsing matchers but it is almost impossible to do so.

I still think it is possible in a limited way. I don't actually need to instrument all subtypes of `org.asynchttpclient.AsyncHandler` but only those which are passed to the `org.asynchttpclient.DefaultAsyncHttpClient#executeRequest(Request request, AsyncHandler<T> handler)` method. So I can just instrument the latter and instrument the exact AsyncHandler subclass which as been provided to the method.
It seems like that would work better by creating an additional `ResettableClassFileTransformer` as opposed to patching an existing one. The disadvantage is that I could end up creating lots of `ResettableClassFileTransformer`s but that's probably not a problem, is it?

But for dynamically reconfiguring the agent this is really useful! Did a POC for that and it works really well 👍

raphw:
Yes about the description strategy even though this is generally true, also for installing. You can migitate the problem by using good caches up to even using a `CachingMatcher` since the class file transformer needs to match the same types once again.

Yes, you can also instrument types delayed once you observe them but this requires you to do a bit of book keeping. In this case, I do however think that patching is the way to go. Alternatively, register a mutable matcher/transformer and call `retransformClasses` manually. As long as your mutations are thread-safe, this should also work. Many class file transformers are not a problem but Byte Buddy would need to resolve transformed types repeatedly if multiple transformers transform the same types. Also, the matching overhead might be noticable as all types are passed through all transformers.

PR #837, opened 2020-04-16
Fix the javadoc for Advice.WithCustomMapping#bind

joshiste:
The used OffsetMapping.Factory only supports strings and primitive
wrappers and not method handles/types as described by the javadoc.
This fixes the javadoc.


raphw:
Oh, that sucks. I think I'll rather fix the support for it in the next release. It's supported everywhere else. Thanks for reporting!

raphw:
Fixed on master.

PR #868, opened 2020-05-19
Add ElementMatchers.takesNoArguments()

joshiste:
For better readability add utility function to match no arguments.


raphw:
I'd say takesArguments(0) works for this?

joshiste:
> I'd say takesArguments(0) works for this?

yes it does. Somehow I stared for 30minutes at all the different overloads until I figured out until how to match no arguments.
So I thought it just might be handy to have a shortcut for this - if you think otherwise (e.g. it will just bloat the API) i'm okay with it.

raphw:
Fair point, does not hurt to add it.

Issue #880, opened 2020-06-12
ClassFormatError when using indy dispatch


felixbarny:
When instrumenting `org.slf4j.LoggerFactory<init>` with a simple enter advice that is dispatched with invokedynamic I get the following exception:

```
java.lang.IllegalStateException: Error invoking java.lang.instrument.Instrumentation#retransformClasses
	at net.bytebuddy.agent.builder.AgentBuilder$RedefinitionStrategy$Dispatcher$ForJava6CapableVm.retransformClasses(AgentBuilder.java:6877) ~[byte-buddy-dep-1.10.11.jar:?]
	at net.bytebuddy.agent.builder.AgentBuilder$RedefinitionStrategy$Collector$ForRetransformation.doApply(AgentBuilder.java:7148) ~[byte-buddy-dep-1.10.11.jar:?]
	at net.bytebuddy.agent.builder.AgentBuilder$RedefinitionStrategy$Collector.apply(AgentBuilder.java:6993) ~[byte-buddy-dep-1.10.11.jar:?]
	at net.bytebuddy.agent.builder.AgentBuilder$RedefinitionStrategy.apply(AgentBuilder.java:4855) ~[byte-buddy-dep-1.10.11.jar:?]
	at net.bytebuddy.agent.builder.AgentBuilder$Default.doInstall(AgentBuilder.java:9463) ~[byte-buddy-dep-1.10.11.jar:?]
	at net.bytebuddy.agent.builder.AgentBuilder$Default.installOn(AgentBuilder.java:9384) ~[byte-buddy-dep-1.10.11.jar:?]
	at net.bytebuddy.agent.builder.AgentBuilder$Default$Delegator.installOn(AgentBuilder.java:10986) ~[byte-buddy-dep-1.10.11.jar:?]
	at co.elastic.apm.agent.bci.ElasticApmAgent.initInstrumentation(ElasticApmAgent.java:173) ~[classes/:?]
	at co.elastic.apm.agent.bci.ElasticApmAgent.initInstrumentation(ElasticApmAgent.java:152) ~[classes/:?]
	at co.elastic.apm.agent.bci.InstrumentationTest.testPatchClassFileVersionToJava7(InstrumentationTest.java:292) ~[test-classes/:?]
	at jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method) ~[?:?]
[...]

PR #882, opened 2020-06-15
Fixes varargs in ForDynamicInvocation.apply

felixbarny:
fixes https://github.com/raphw/byte-buddy/issues/880#issuecomment-643974973


felixbarny:
too slow 😄

Issue #820, opened 2020-03-10
Add metadata necessary to be recognized by m2e out of the box


odrotbohm:
It would be cool if the ByteBuddy Maven plugin would automatically be detected and properly executed within m2e (Eclipse Maven integration). Ideally that would also cause execution errors to properly reported in Eclipse's problems view etc.

Currently m2e is not able to detect proper lifecycle mappings (see [the docs](https://www.eclipse.org/m2e/documentation/m2e-execution-not-covered.html#lifecycle-mapping-metadata-provided-by-maven-plugin))but you can Eclipse to invoke the plugin by editing the lifecycle metadata configuration (Settings > Maven > Lifecycle Mappings > Open workspace lifecycle mappings metadata) and adding the following snippet of XML.

```xml
<pluginExecution>
  <pluginExecutionFilter>
    <groupId>net.bytebuddy</groupId>
    <artifactId>byte-buddy-maven-plugin</artifactId>
    <versionRange>[0.0.1,)</versionRange>
    <goals>
      <goal>transform</goal>
    </goals>
  </pluginExecutionFilter>
  <action>
    <execute>
      <runOnIncremental>true</runOnIncremental>
    </execute>
  </action>
</pluginExecution>
```

The m2e developer documentation contains [useful hints](https://www.eclipse.org/m2e/documentation/m2e-making-maven-plugins-compat.html) on what it takes to make this work.

raphw:
Fixed bt #897.

odrotbohm:
Looks like this has been resolved and published already? According to Maven Central 1.10.13 is already out.

odrotbohm:
I have to take that back. After a M2E configuration cleanups and a restart I am back to where I started. A bit of investigation revealed that the current [version of the metadata file](https://github.com/raphw/byte-buddy/blob/master/byte-buddy-maven-plugin/src/main/resources/META-INF/m2e/lifecycle-mapping-metadata.xml) simply contains:

```xml
<action>
  <execute />
</action>
```

whereas the [developer documentation](https://www.eclipse.org/m2e/documentation/m2e-making-maven-plugins-compat.html) states that `<runOnIncremental>true</runOnIncremental>` would be needed to execute the plugin during Eclipse builds.

[...]

PR #897, opened 2020-07-08
Add resource directory to resources

### LINKED ISSUE #896: maven-plugin Eclipse Metadata not published
The required metadata for eclipse is under:
 https://github.com/raphw/byte-buddy/tree/master/byte-buddy-maven-plugin/src/main/resources/META-INF/m2e
but is it not included in 
https://repo1.maven.org/maven2/net/bytebuddy/byte-buddy-maven-plugin/1.10.13/byte-buddy-maven-plugin-1.10.13.jar
the released jar
I think the problem is the Filtering in the pom.xml of the maven plugin:

```
<build>
        <resources>
            <resource>
                <directory>..</directory>
                <targetPath>META-INF</targetPath>
                <filtering>true</filtering>
                <includes>
                    <include>LICENSE</include>
                    <include>NOTICE</include>
                </includes>
            </resource>
        </resources>
    </build>
```
This should also include META-INF/m2e

tkalmar:
Fixes #896 and #820

### DIFF
--- a/byte-buddy-maven-plugin/pom.xml
+++ b/byte-buddy-maven-plugin/pom.xml
@@ -93,6 +93,9 @@
     <build>
         <resources>
             <resource>
+                <directory>src/main/resources</directory>
+            </resource>
+            <resource>
                 <directory>..</directory>
                 <targetPath>META-INF</targetPath>
                 <filtering>true</filtering>


raphw:
Thanks, good catch!

PR #918, opened 2020-09-01
move evaluation of `byteBuddyExtension.implies` into TransformationAc…

rmlekus:
The Project extension "ByteBuddyExtension" is handled by "lazy configuration", thus it not yet initialized when registering the PostCompilationAction. (see https://docs.gradle.org/current/userguide/lazy_configuration.html) thus the check with-> byteBuddyExtension.implies(task) check will encounter a null value in the tasks set, and actually register the TransformationAction always.

To fix it, I moved `byteBuddyExtension.implies(task)` from PostCompilationAction.execute to TransformationAction.apply.

Configuring an example project with:
```groovy
byteBuddy {
  transformation {
    plugin = "******Plugin"
  }
  tasks = ['compileJava'].toSet()
}
```
thereafter worked as expected.


raphw:
Thanks for your improvement, good catch. In favor of the new Gradle plugin on the branch *incremental-build*, this change will become obsolete, unfortunately. The new plugin defines its targets based on source set-specific tasks.

PR #955, opened 2020-10-14
Fixes #954 by making the assignability check more correct

### LINKED ISSUE #954: The type assignability check in JavaConstant.Dynamic.ofInvocation() is too strict
Hi; it's me again; sorry!

For background: https://stackoverflow.com/questions/64360249/is-type-assignability-too-strict-in-javaconstant-dynamic-ofinvocation

In very concrete terms, I think that this assignability check:

https://github.com/raphw/byte-buddy/blob/e4ce4af1bab578a7ac3497d07449e50628d050e7/byte-buddy-dep/src/main/java/net/bytebuddy/utility/JavaConstant.java#L1636-L1638

…is too strict given that it prevents a method from being invoked that uses varargs.  Specifically, `iterator.next()` will, in such a case, return something like `Glorp[]`, and (let's say) `Glorp` is not assignable to `Glorp[]`.

Suppose I have:
```
static final JavaConstant toJavaConstant(final Glorp[] glorps) {
  final JavaConstant[] javaConstants = new JavaConstant[glorps.length];
  for (int i = 0; i < javaConstants.length; i++) {
    javaConstants[i] = toJavaConstant(glorps[i]); // another version of this method that works on scalars
  }
  return JavaConstant.Dynamic.ofInvocation(SOME_METHOD_THAT_ACCEPTS_A_VARARGS_OF_THINGS, javaConstants);
}
```
…and `SOME_METHOD_THAT_ACCEPTS_A_VARARGS_OF_THINGS` is a `MethodDescription.InDefinedShape` that resolves to something like this:
```
public static final Glorp[] toGlorps(final Glorp... glorps) {
  return glorps;
}
```
Currently, ByteBuddy will not allow this because of that type assignability check.

If I comment out lines 1636 through 1638, i.e. if I disable the type assignability check, this construction works just fine (probably thanks to all the `MethodHandle` argument conversion machinery happening at constant resolution time).

I'll file a PR shortly.

ljnelson:

### DIFF
--- a/byte-buddy-dep/src/main/java/net/bytebuddy/utility/JavaConstant.java
+++ b/byte-buddy-dep/src/main/java/net/bytebuddy/utility/JavaConstant.java
@@ -1633,8 +1633,11 @@ public interface JavaConstant {
                         throw new IllegalArgumentException("Not a compile-time constant: " + constant);
                     }
                 }
-                if (!typeDescription.isAssignableTo(iterator.next())) {
-                    throw new IllegalArgumentException("Cannot assign " + constants + " to " + methodDescription);
+                TypeDescription next = iterator.next();
+                if (!typeDescription.isAssignableTo(next)) {
+                    if (!methodDescription.isVarArgs() || iterator.hasNext() || !next.isArray() || !typeDescription.isAssignableTo(next.getComponentType())) {
+                        throw new IllegalArgumentException("Cannot assign " + constants + " to " + methodDescription);
+                    }
                 }
             }
             return new Dynamic(new ConstantDynamic("invoke",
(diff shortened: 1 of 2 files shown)


ljnelson:
Hi, @raphw; I found an issue in `JavaConstant.Dynamic#ofInvocation(MethodDescription.Invocation)` (#954).  I fixed it here (and added a unit test).

ljnelson:
(Not sure why all the actions are failing; a full build passes all tests locally.)

ljnelson:
Oh, this is interesting.  JDK 15 (what I was using) I guess is better at type inference than 11.  That very same Hamcrest assertion works just fine under later JDKs.  Let me see if I can coax JDK 11 into accepting it.

PR #962, opened 2020-10-18
Fixes #961 by handing JavaConstant case in FixedValue.value() method

### LINKED ISSUE #961: FixedValue#value(Object) does not consider JavaConstant
I'm aware that `FixedValue` has two `value` methods.  One takes a `JavaConstant` and the other takes an `Object`.

The one that takes an `Object` does not check to see if the `Object` is, itself, a `JavaConstant` implementation.  I think it should, and should then delegate to the `value(JavaConstant)` method.

(Obviously the work around is for the user herself to do the (tedious, boring) `instanceof` check before choosing which method to invoke.)

I'll file a PR soon.

ljnelson:

### DIFF
--- a/byte-buddy-dep/src/main/java/net/bytebuddy/implementation/FixedValue.java
+++ b/byte-buddy-dep/src/main/java/net/bytebuddy/implementation/FixedValue.java
@@ -70,9 +70,9 @@ public abstract class FixedValue implements Implementation {
     /**
      * <p>
      * Returns a fixed value from any intercepted method. The fixed value is stored in the constant pool if this is possible.
-     * Java is capable of storing any primitive value, {@link String} values and {@link Class} references in the constant pool.
-     * Since Java 7, {@code MethodHandle} as well as {@code MethodType} references are also supported. Alternatively, the fixed
-     * value is stored in a static field.
+     * Specifically, an argument that is a {@link JavaConstant}, {@link TypeDescription}, primitive, {@link String} or
+     * {@link Class} value is stored in the constant pool.  Since Java 7, {@code MethodHandle} as well as {@code MethodType}
+     * references are also supported. Alternatively, the fixed value is stored in a static field.
      * </p>
      * <p>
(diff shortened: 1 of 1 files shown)


ljnelson:
Whoops; let me add `TypeDescription` in here too.

ljnelson:
OK; @raphw see what you think

raphw:
I'm not fully understanding the first scenario where the value itself is returned, is this in case you accidentally pass the implementation itself? I'm not sure if I'd like to support this, sounds like an edge case that only happens if there's an error in the code already.

ljnelson:
OK, the thought was just: if you accept an `Object`, then your contract says you accept _anything_.

I'll look at the javadoc.

ljnelson:
Which part of the javadoc did you need fixed?  Do you want me to remove the `@see` tags?

raphw:
Well, `Object` is the most specific common type of all arguments, it's more of a convenience option to intrinsify a choice that a user would otherwise need to take. The contract is in the javadoc which should mention what values are acceptable.

ljnelson:
Sounds like you would prefer if I removed the `if (fixedValue instanceof AssignerConfigurable)` check.  Would that solve the issues?

raphw:
Yes, if you remove it and adjust the javadoc, I'll merge and make a new release.

ljnelson:
OK.  What javadoc adjustments are needed?  All I did was add `@see` tags.  Would you prefer those be removed?

ljnelson:
OK, I've adjusted the javadocs (did not remove the `@see` tags but made it more clear, I hope, about what arguments are acceptable.  I also removed the `AssignerConfigurable` check.  See what you think.

ljnelson:
One hopefully last question.  I just noticed while looking through `MethodCall.ArgumentLoader.ForStackManipulation#of()` that it doesn't deal with `TypeDescription`s.  Should it as well?  And `JavaConstant`?  (I see that it deals with `Class`.)

PR #965, opened 2020-10-23
Fix class injection for legacy classloaders which use getPackage internally on JDK9+

agoallikmaa:
The old behavior of checking if a package is already defined in `ClassLoader#definePackage` was to rely on `getPackage` which asks the parent classloader if it was not found to be defined in the current one. This changed with JDK9 - now it is checked only against packages defined in the same classloader.

However, some custom classloaders (encountered the issue on WLS14+JDK11 combination, in its `GenericClassLoader`) which were originally written for JDK<9 still override the `definePackage` with the legacy behavior. This makes the `getDefinedPackage` check return null, but `definePackage` will still fail because it finds that the package already exists via `getPackage`.

In this case, the only recourse seems to be to consider this case as a success - call `getPackage` to verify the package exists in an ancestor and then check that it is compatible. Unfortunately this means that both `getDefinedPackage` and `getPackage` need to be available, meaning one extra field in several classes.

Could not find a reason for the `getPackage.setAccessible` to be called in `UsingUnsafeInjection.make`. It seems like it was unnecessary as that `Method` instance is not directly invoked, the method is only accessed through the generated accessor, and it is not done there for other methods with identical access modifiers.


raphw:
Thanks, good catch!

PR #981, opened 2020-11-25
Tweak m2e configuration to let the Maven plugin run on incremental builds as well.

odrotbohm:
Prior to this change, when editing sources in Eclipse and saving the source file, the transformations registered by the plugin would not run and thus leave the classes in invalid state until a complete project refresh was triggered.

### DIFF
--- a/byte-buddy-maven-plugin/pom.xml
+++ b/byte-buddy-maven-plugin/pom.xml
@@ -55,6 +55,11 @@
             <version>${version.maven.aether}</version>
         </dependency>
         <dependency>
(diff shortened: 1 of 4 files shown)


raphw:
This looks promising. I am reworking it a bit but I am certain this will make it to master. The existence of this API is new to me so I want to read up a bit before committing. I will also add a switch for disabling incremental builds and likely make it the default but I need to understand how this interacts with the Eclipse plugin, too.

raphw:
Already merged directly.

PR #995, opened 2021-01-04
Matcher deduplication

richardstartin:
This is small effort towards reducing the footprint of the matching system. Since none of these matchers are data dependent, they can be singletons. In the Datadog agent with default settings (as a concrete use case), during startup we allocate 27768B of `MethodSortMatcher` and 11328B of `ModifierMatcher` during bootstrapping. `BooleanMatcher` accounts for another 17KB. Allocating all of these once when the `Sort` and `Mode` enums are loaded eliminates this. It's not a huge impact but I think it's a worthwhile change, has no downsides, and composes additively with any other footprint improvements. 

Deduplicates most static cardinality matchers
* `ModifierMatcher`
* `MethodSortMatcher`
* `BooleanMatcher`
* `NullMatcher`

### DIFF
--- a/byte-buddy-dep/src/main/java/net/bytebuddy/matcher/MethodSortMatcher.java
+++ b/byte-buddy-dep/src/main/java/net/bytebuddy/matcher/MethodSortMatcher.java
@@ -28,6 +28,16 @@ import net.bytebuddy.description.method.MethodDescription;
 public class MethodSortMatcher<T extends MethodDescription> extends ElementMatcher.Junction.AbstractBase<T> {
 
     /**
+     * Returns an element matcher that matches a specific sort of method description.
+     *
+     * @param sort The sort of method description to be matched by this element matcher.
+     */
+    @SuppressWarnings("unchecked")
+    public static <T extends MethodDescription> MethodSortMatcher<T> of(Sort sort) {
+        return (MethodSortMatcher<T>) sort.matcher;
+    }
+
+    /**
      * The sort of method description to be matched by this element matcher.
      */
     private final Sort sort;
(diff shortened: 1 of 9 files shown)


byte-buddy-dep/src/main/java/net/bytebuddy/matcher/ModifierMatcher.java
```diff
@@ -167,6 +182,7 @@ public String toString() {
         Mode(int modifiers, String description) {
             this.modifiers = modifiers;
             this.description = description;
+            this.matcher = new ModifierMatcher<ModifierReviewable>(this);
```
richardstartin, line 185: This isn't unsafe publication in this particular case because it's an enum.

byte-buddy-dep/src/main/java/net/bytebuddy/matcher/MethodSortMatcher.java
```diff
@@ -114,13 +124,16 @@ protected boolean isSort(MethodDescription target) {
...
+        private final MethodSortMatcher<?> matcher;
+
         /**
          * Creates a new method sort representation.
          *
          * @param description A textual representation of the method sort that is represented by this instance.
          */
         Sort(String description) {
             this.description = description;
+            this.matcher = new MethodSortMatcher<MethodDescription>(this);
```
richardstartin, line 136: This isn't unsafe publication in this particular case because it's an enum.

richardstartin:
Apologies for the previously obviously broken third commit - since it seems I am the first this year, `mvn clean verify` updates the copyright notices so I have been avoiding running the whole suite locally.

raphw:
All good, this is a very sensible change. The matchers are used a lot so reducing them is a good addition.

PR #1004, opened 2021-01-28
Documentation adjustment for Java HotSwap feature

adamantal:
Hi!

I happen to browse the documentation today and following the link you've provided in section `Reloading a class` in the `Tutorial` page it seems that OpenJDK has given up on extending HotSwap functionality. See [JDK-8046149](https://bugs.openjdk.java.net/browse/JDK-8046149) for the issue. They haven't actually provided explanation in that ticket, so the actual reason why they decided to close this issue may lie in the JDK email list.
Anyways, I think it would be useful to mention this in Byte Buddy's documentation.


raphw:
I missed that. Bit of a shame but not unexpected. Thanks for the update!

Issue #1036, opened 2021-04-13
Collapsing nesting conjunctions and disjunctions


richardstartin:
In the Datadog tracer we often see stack traces like this where there are multiple layered conjunctions or disjunctions

<img width="929" alt="Screenshot 2021-04-13 at 10 50 17" src="https://user-images.githubusercontent.com/16439049/114533793-3245a180-9c46-11eb-99e0-eb5c2c0c1663.png">

I expect that inlining these binary matchers into a wider N-way conjunction/disjunction matcher would be profitable (but haven't actually done any experiments yet). This would be straightforward to do eagerly without requiring an optimising pass over the tree. 

Currently the matchers have a one to one correspondence with what the user defines, and care has been taken to represent the structure faithfully in the `toString` implementations, but is this something you have previously thought about doing? Is there any potential to break something I haven't thought about or other users rely on/expect?

raphw:
That's a good point and I considered it from time to time, mainly for the `ElementMatcher` API but it's just as legitimate for the agent-related matchers. I always postponed it for the binary incompatible change but now is as good of a time as ever for a new feature release.

I fixed it in the most compatible manner possible and applied it for all matchers, it's not a big change so it was easy enough. Thanks for pushing this back into my memory.

PR #1037, opened 2021-04-13
collapse logical conjunction and disjunction raw matchers

richardstartin:
This is more to add colour to #1036 than submitted with an expectation to merge. Since the collapsing logic is the same for conjunctions and disjunctions, I merged the classes and differentiate them with an operation identifier.

### DIFF
--- a/byte-buddy-dep/src/main/java/net/bytebuddy/agent/builder/AgentBuilder.java
+++ b/byte-buddy-dep/src/main/java/net/bytebuddy/agent/builder/AgentBuilder.java
@@ -1525,21 +1525,29 @@ public interface AgentBuilder {
             }
         }
 
+
         /**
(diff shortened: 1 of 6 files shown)


richardstartin:
This would also allow for keeping an array of counters in the matcher and tracking how often matchers return false and reordering the decomposed matchers. For conjunctions, these should be evaluated early to get out quicker. For disjunctions they should be pushed to the back of the list.

richardstartin:
Is the HotSpot (legacy) failure related to this change?

raphw:
The failure is related to some bug in Java 6 for the OpenJ9 emulation test. I tried to reproduce it to avoid it, maybe I'll skip the test on Java 6 eventually. I fixed this in a slightly different way, therefore I am closing this PR.

PR #1041, opened 2021-04-19
Use a singleton instance for MethodVariableAccess.loadThis()

rovarga:
MethodVariableAccess.loadThis() is a rather common instruction, which unfortunately ends up costing an object for each invocation. Use a lazily-instantiated singleton object instead.


raphw:
Thanks, good point. I have a declarative mechanism for this which I prefer, that's why I won't merge it but will include it in the next release.

PR #1048, opened 2021-05-10
StackManipulation op expansion

Firestar99:
A collection of small additions I've made for my own project or wanted to have. Sending them upstream for everyone else to enjoy!

834c242 9fb1087 I needed some additional StackManipulation operations specifically Divide, Remainder and ArrayLength

10a71c4 b032050 Sometimes you type TextConstant expecting there to be a of() method like everywhere else but there isn't. You need to new TextConstant it. So I've gone though and added the missing ones. b032050 also adds of() methods for Singletons without any arguments, thought I'd split that in case that is not wanted or needed.

0bc6dd7  PrimitiveWideningDelegate is used usually as `PrimitiveWideningDelegate.INTEGER.widenTo(ForLoadedType.of(long.class))` and I wanted to short that to `PrimitiveWideningDelegate.INTEGER.widenToLong()` (needed to do lots of I2L casts in my project)

Hope these are helpful, love Bytebuddy for writing code that writes code :)

### DIFF
--- a/byte-buddy-dep/src/main/java/net/bytebuddy/implementation/bytecode/Division.java
+++ b/byte-buddy-dep/src/main/java/net/bytebuddy/implementation/bytecode/Division.java
@@ -20,29 +20,29 @@ import org.objectweb.asm.MethodVisitor;
 import org.objectweb.asm.Opcodes;
 
 /**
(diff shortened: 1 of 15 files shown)


Firestar99:
also here some tests for the new StackManipulations :)

Issue #1059, opened 2021-06-02
Problem with array-typed ClassDescs in JavaConstant.java leads to incorrect descriptor string somewhere


ljnelson:
I apologize for the vague issue title; there's a lot going on here.

Consider:

https://github.com/raphw/byte-buddy/blob/77fbfb937d376260ac5400028c2942ac9f0f390d/byte-buddy-dep/src/main/java/net/bytebuddy/utility/JavaConstant.java#L292-L329

Within this range of code, there are I think ten (or so) instances of `net.bytebuddy.jar.asm.Type.getClassName()` being passed to `TypePool.describe()`.  When this class name is, for example, `java.lang.reflect.Type[]` (my particular arbitrary case), the lazy type description that gets created will not in fact be able to be resolved.

More specifically, consider this method when `name` is "`java.lang.reflect.Type[]`" (again, my arbitrary example, encountered during debugging):

https://github.com/raphw/byte-buddy/blob/77fbfb937d376260ac5400028c2942ac9f0f390d/byte-buddy-dep/src/main/java/net/bytebuddy/pool/TypePool.java#L507-L530

Specifically, at line 512 "`java.lang.reflect.Type[]`" will _not_ start with `[` because `name` in this case is not a descriptor string.

I don't know much more than this but I do know that this problem leads to a `ClassFormatError` if a ByteBuddy-generated type makes use of a constant generated in this manner.

I will add more details to this issue if I can find them.

ljnelson:
So how, you may ask, did "`java.lang.reflect.Type[]`" end up as the name in question?  It's what is returned by `java.lang.reflect.Type.class.describeConstable().orElseThrow().arrayType().displayName()`.  I'm not familiar enough with the ByteBuddy dispatching pattern (for old VMs) to know how this got used, but maybe `java.lang.reflect.Type.class.describeConstable().orElseThrow().arrayType().descriptorString()` needs to be used instead?

ljnelson:
I tried replacing `getClassName()` with `getDescriptor()` just quickly, but `TypePool::describe` fails in this case because the descriptor returned contains `/` characters.

ljnelson:
See #1060.  This fixes the problem but may not be where you want to solve it; not sure.

raphw:
[...]

PR #1060, opened 2021-06-03
Makes TypePool::describe accept more inputs

ljnelson:
Fixes issue #1059. 


raphw:
I want the input to stay consistent and only accept the value of `java.lang.Class.getName()`. ASM produces a rather synthetic value that is not used anywhere else. I addressed this issue by adjusting the input.

PR #1062, opened 2021-06-04
Reverts 977b809b0e1c7c99ea0d7a4a532bbc27cd650500 to fix #1061

### LINKED ISSUE #1061: MethodCall.MethodLocator.ForElementMatcher::resolve is now too restrictive
(I hope I have the class name right!)

48698d9d7d49aa2d939bdf4a93cce2d4f643a27b and 977b809b0e1c7c99ea0d7a4a532bbc27cd650500 exclude the instrumented method from element matching.

But there is at least one common case where this is too restrictive: a [certain kind of (common) proxy class](https://stackoverflow.com/questions/67827844/elementmatchersistostring-fails-in-bytebuddy-1-11-1-when-tostring-is-not-def).

Consider a class:
```
public class Z {
  @Override // Object
  public String toString() { return "Z"; }
}
```
Now consider a class that ByteBuddy might generate that, if it were written in Java, would look something like this:
```
public class Proxy$Z extends Z {
  private final Z z;
  public Proxy$Z(Z z) {
    this.z = z;
  }
  private Z proxied() {
    return this.z;
  }
  @Override // Z
  public String toString() {
    return this.proxied().toString();
  }
}
```
To build this using ByteBuddy, we might do the usual `DynamicType.Builder` setup stuff, and then:
```
builder = builder
  .method(isToString()) // doesn't match in 1.11.1
  .intercept(MethodCall.invoke(named("toString"))
             .onMethodCall(MethodCall.invoke(named("proxied"))))
```
This case is prohibited in ByteBuddy 1.11.1 because `resolve` now excludes the instrumented method (`Z.toString()`) from the list of candidates.

I am not sure if both commits need to be reverted/changed or just one of them.

ljnelson:

### DIFF
--- a/byte-buddy-dep/src/main/java/net/bytebuddy/implementation/MethodCall.java
+++ b/byte-buddy-dep/src/main/java/net/bytebuddy/implementation/MethodCall.java
@@ -808,7 +808,7 @@ public class MethodCall implements Implementation.Composable {
                 List<MethodDescription> candidates = CompoundList.<MethodDescription>of(
                         instrumentedType.getSuperClass().getDeclaredMethods().filter(isConstructor().and(matcher)),
                         instrumentedType.getDeclaredMethods().filter(not(ElementMatchers.isVirtual()).and(not(is(instrumentedMethod))).and(matcher)),
-                        methodGraphCompiler.compile(targetType, instrumentedType).listNodes().asMethodList().filter(not(is(instrumentedMethod)).and(matcher)));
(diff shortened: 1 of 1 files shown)


ljnelson:
With this PR, the issue described in #1061 is fixed.  Please do have a close look as I'm not sure of all the repercussions.

raphw:
Sorry for not merging, I added some tests and the change was so small that I did not base it upon it. Thanks for the investigation and report!

Issue #1069, opened 2021-06-17
InvokeDynamic.TerminationHandler.DROPPING uses wrong return type in chaining situations


ljnelson:
A [StackOverflow question describing this issue](https://stackoverflow.com/questions/68008319/can-i-compose-an-invokedynamic-with-a-fixedvalue-using-implementation-composable) has more details.

Briefly:

When building up an `Implementation.Composable` consisting of (minimally) an `InvokeDynamic` followed by a `FixedValue`, the `TerminationHandler.DROPPING` instance responsible for dropping the return type of the `InvokeDynamic` operation wrongly drops the return type of the intercepted method, not the return type of the `InvokeDynamic` operation.

That is, with this:
```
// assume we're implementing a method defined as: public String foo(String arg0);
// assume the InvokeDynamic instruction here calls the equivalent of public void frob();
// note the difference in return types
composableImplementation = InvokeDynamic.bootstrap(...).invoke(...).withArgument(0); // pretend the return type is void
composableImplementation = composableImplementation.andThen(FixedValue.argument(0)); // pretend argument 0 is String
```
...the termination handler "in between" the `InvokeDynamic` and the `FixedValue` will attempt to pop a `String`—the return type of the intercepted method `foo()`—when it should be trying to pop a `void` (the return type of the `invokedynamic` instruction).

Changing the code found here:

https://github.com/raphw/byte-buddy/blob/fef2ef91ae9778010c582180190dbde86da9ca78/byte-buddy-dep/src/main/java/net/bytebuddy/implementation/InvokeDynamic.java#L2485-L2495

...so that `Removal.of()` accepts `returnType`, like so:
```
return Removal.of(returnType);
```
…fixes the problem.

ljnelson:
Additionally, I think the assumption in this method (and all of the related "convenience" methods):

https://github.com/raphw/byte-buddy/blob/fef2ef91ae9778010c582180190dbde86da9ca78/byte-buddy-dep/src/main/java/net/bytebuddy/implementation/InvokeDynamic.java#L2905-L2925

…is incorrect.

The documentation says:
> The return type is he [sic] assigned to the intercepted method's return type.

I think without the typos what this means to say is: "The return type must be assignable to the intercepted method's return type."  But I don't think this is true.

[...]

PR #1070, opened 2021-06-17
Fixed TerminationHandler.DROPPING to drop the proper return type

ljnelson:


ljnelson:
This causes the problems encountered as described in #1069 to be resolved.  I make no claim here that this is the proper fix, but it is _a_ fix. 😄

PR #1079, opened 2021-07-06
Add missing entries to module-info.java

yrodiere:
If I understand correctly the commit message of d3993051f2efcb0cef5523e1638d4dd2bfe743d8 , the package `net.bytebuddy.utility.dispatcher` is API and thus should be exposed to other modules.

Regardless, if it isn't exposed, I get errors in Hibernate ORM tests when running tests in the module paths; see https://github.com/hibernate/hibernate-orm/pull/4082#issuecomment-874172474.

### DIFF
--- a/byte-buddy/pom.xml
+++ b/byte-buddy/pom.xml
@@ -60,6 +60,7 @@
             net.bytebuddy.matcher,
             net.bytebuddy.pool,
             net.bytebuddy.utility,
+            net.bytebuddy.utility.dispatcher,
             net.bytebuddy.utility.privilege,
             net.bytebuddy.utility.visitor,
             ${shade.target},


yrodiere:
I think the failing test is unrelated; the logs are not helpful: https://pipelines.actions.githubusercontent.com/POiHbLMS71E4NiMGhicPNxkPyU0P1YU1ziMnM3szgYnTWbH72H/_apis/pipelines/1/runs/692/signedlogcontent/18?urlExpires=2021-07-06T09%3A07%3A12.1722481Z&urlSigningMethod=HMACV1&urlSignature=kCVdKMmxiTIrBRmjEz4lLFDmwhAodkarmo1F8KstkvA%3D

raphw:
The idea was that the dispatcher remains internal API to modules as it is security sensitive in a module context. Do you understand the error?

yrodiere:
> Do you understand the error?

[This one?](https://github.com/hibernate/hibernate-orm/pull/4082#issuecomment-874561406) Only superficially. Somehow the `JavaDispatcher` class is not available in the classloader of the `TypeDescription$ForLoadedType` class. Why would that be... that I do not understand.

> I don't think its internalization should make a difference to an API user?

Neither do I. In theory this class should be accessible from classes of its own module. But here we are.

I suspect there is a classloader issue in ORM, but I was hoping I could delay the investigation by just fixing your `module-info.java`. Since the fix is invalid... I'll get to work :/

raphw:
I found the issue on my end. Planning a fix release tonight. I'll see how I can also add an automated test.

raphw:
I found the issue, a *module-info* must mention any package that it considers, otherwise the module class loader will not even attempt to load a class starting from the module path. Since I still support Java 5, I wrote a custom Maven plugin to create the *module-info*. I forgot to add the new package, and this caused the `NoClassDefFoundError`. Could you try version *1.11.7* which I just released? It should resolve this issue.

yrodiere:
The dispatcher class is found now, but there is definitely a classloader issue, probably caused by Hibernate ORM's "aggregated" classloader:

```
java.lang.IllegalAccessError: superinterface check failed: class net.bytebuddy.utility.dispatcher.JavaDispatcher$Invoker$Dispatcher (in unnamed module @0x266c159d) cannot access class net.bytebuddy.utility.dispatcher.JavaDispatcher$Invoker (in module net.bytebuddy) because module net.bytebuddy does not export net.bytebuddy.utility.dispatcher to unnamed module @0x266c159d
	at java.base/java.lang.ClassLoader.defineClass1(Native Method)
	at java.base/java.lang.ClassLoader.defineClass(ClassLoader.java:1017)
	at net.bytebuddy@1.11.7/net.bytebuddy.utility.dispatcher.JavaDispatcher$DynamicClassLoader.invoker(JavaDispatcher.java:1307)
	at net.bytebuddy@1.11.7/net.bytebuddy.utility.dispatcher.JavaDispatcher$Invoker$CreationAction.run(JavaDispatcher.java:443)
	at net.bytebuddy@1.11.7/net.bytebuddy.utility.dispatcher.JavaDispatcher$Invoker$CreationAction.run(JavaDispatcher.java:432)
	at java.base/java.security.AccessController.doPrivileged(Native Method)
	at net.bytebuddy@1.11.7/net.bytebuddy.utility.dispatcher.JavaDispatcher.doPrivileged(JavaDispatcher.java)
	at net.bytebuddy@1.11.7/net.bytebuddy.utility.dispatcher.JavaDispatcher.<clinit>(JavaDispatcher.java:65)
	at net.bytebuddy@1.11.7/net.bytebuddy.description.type.TypeDescription$ForLoadedType.<clinit>(TypeDescription.java:8378)
	at net.bytebuddy@1.11.7/net.bytebuddy.description.type.TypeDescription.<clinit>(TypeDescription.java:65)
	at net.bytebuddy@1.11.7/net.bytebuddy.matcher.ElementMatchers.isFinalizer(ElementMatchers.java:1626)
	at net.bytebuddy@1.11.7/net.bytebuddy.matcher.ElementMatchers.isDefaultFinalizer(ElementMatchers.java:1616)
	at net.bytebuddy@1.11.7/net.bytebuddy.ByteBuddy.<init>(ByteBuddy.java:207)
	at org.hibernate.orm.core@5.5.4-SNAPSHOT/org.hibernate.bytecode.internal.bytebuddy.ByteBuddyState.<init>(ByteBuddyState.java:79)
```

raphw:
No, I think I understand what this is. I will try another test and fix for this.

yrodiere:
Ok. If you need me to test another patch, just tell me; I can definitely build Byte Buddy locally in order to test it.

raphw:
This time I did a test that actually used the module system instead of pretending to, and it works. I struggle a bit to automate such a test since I inject the module-info but I keep it in the back of my head.

I am now releasing 1.11.8 to hopefully get past this issue for good now.

yrodiere:
The ORM tests seem to pass with Byte Buddy 1.11.8 in the modulepath. I'll let the full test suite run and will report back if there are other problems, but I think we're good now :)

Thanks a lot @raphw!

raphw:
Great, many thanks to you. Those module problems can be rather cryptic. I'll see if I can better automate this.

PR #1093, opened 2021-08-13
Fix typo

perlun:
Found this when looking at the Javadocs, gave it a look and found a few instances of the same typo in a couple of classes.

(Thanks for a _great_ Java library! This is my way to say "thank you" for all the awesome work you've done and continue doing over the years. :pray:)


raphw:
Much appreciated! Glad you like the library.

PR #1119, opened 2021-09-27
Add AssignTo annotations to simpify assignments for non-inlined advices

felixbarny:
The `readOnly` property can only be set to `false` in inlined advices (`inline=true` (default value)). Similarly, `@Local` can only be used in inlined advices.

This makes is quite tricky to work with non-inlined advices where Byte Buddy delegates to the advice via a static method call or via an `INVOKEDYNAMIC` instruction.

For the Elastic APM Java agent, we have created a mechanism that uses the return value of an advice method to assign it to things like the instrumented method's return value, arguments, fields, or the thrown exception. To support multiple assignments, the advice can return an `Object[]`

Example:

```java
@AssignTo(arguments = {
    @AssignTo.Argument(index = 0, value = 1),
    @AssignTo.Argument(index = 1, value = 0)
})
@Advice.OnMethodEnter(inline = false)
public static Object[] onEnter(@Advice.Argument(0) String foo, @Advice.Argument(1) String bar) {
    return new Object[]{foo, bar};
}
```

While this causes more allocations (for the `Object[]`), there's a good chance for the JIT's escape analysis to replace it with a stack allocation.

To avoid accidentally assigning a `null` value when exceptions are suppressed, the `@AssignTo` annotations ignore `null` values.

```java
@AssignTo.Return
@Advice.OnMethodExit(suppress = Throwable.class, onThrowable = Throwable.class, inline = false)
public static String onMethodExit(@Advice.Thrown Throwable throwable) {
    throw new RuntimeException();
}
```

This can come as a surprise when explicitly wanting to assign a null value, for example to suppress an exception in the instrumented method.

```java
@AssignTo.Thrown
@Advice.OnMethodExit(onThrowable = Throwable.class, inline = false)
public static Throwable onMethodExit(@Advice.Thrown Throwable throwable) {
    return null;
}
```

To work around that, users have to return an `Object[]` instead.
```java
@AssignTo.Thrown(index = 0)
@Advice.OnMethodExit(onThrowable = Throwable.class, inline = false)
public static Object[] onMethodExit(@Advice.Thrown Throwable throwable) {
    return new Object[]{null};
}
```

Maybe there are better strategies for handling `null` values and suppressed exceptions, such as only skipping the `null` assignment if there actually was a `suppressed` exception.

### DIFF
--- a/byte-buddy-dep/src/main/java/net/bytebuddy/asm/Advice.java
+++ b/byte-buddy-dep/src/main/java/net/bytebuddy/asm/Advice.java
@@ -20,6 +20,7 @@ import net.bytebuddy.ClassFileVersion;
 import net.bytebuddy.build.HashCodeAndEqualsPlugin;
 import net.bytebuddy.description.annotation.AnnotationDescription;
 import net.bytebuddy.description.annotation.AnnotationValue;
+import net.bytebuddy.description.annotation.AnnotationList;
 import net.bytebuddy.description.enumeration.EnumerationDescription;
 import net.bytebuddy.description.field.FieldDescription;
 import net.bytebuddy.description.method.MethodDescription;
@@ -66,6 +67,7 @@ import java.lang.reflect.Field;
 import java.lang.reflect.Method;
 import java.util.*;
 
+import static net.bytebuddy.matcher.ElementMatchers.annotationType;
 import static net.bytebuddy.matcher.ElementMatchers.*;
 
 /**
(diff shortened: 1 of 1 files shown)


felixbarny:
An entirely different approach would by trying to support `readOnly=false` and `@Local` for non-inlined advices by transparently mapping assignments to an `Object[]` and back. Not sure if that's feasible, though.

raphw:
I am playing with the thought to rather of overloading the `@Enter` value, I'd introduce a new binder in the style of:

```java
@Advice.OnMethodEnter(inline = false)
public static void onEnter(
  @Advice.Argument(0) String foo, 
  @Advice.Argument(1) String bar,
  @Advice.AssignTo(arguments = {
    @Advice.AssignTo.Argument(index = 0, value = 1),
    @Advice.AssignTo.Argument(index = 1, value = 0)
  }) Object[] assignment
) {
    assignment[0] = "foo";
    assignment[1] = "bar";
}
```

As you say, I think that escape analysis within the method will likely avoid the boxing for sufficiently short methods. My idea now for avoiding the exception problem is that I'd create a new value `new Object()` and use this as the identity for not assigning a value. So the example would be implemented as:

```java
Object marker = new Object();
Object[] markers = new Object[] {marker, marker};
onEnter(foo, bar, markers);
if (markers[0] != marker) { ... }
if (markers[1] != marker) { ... }
```

I think that EA would still do the job, one could still use the enter mechanism as expected and there would be a cleaner divide as a result. Any thoughts on this approach?

felixbarny:
Hmm, interesting idea.
Maybe I have become too paranoid about allocations at this point but the advantage of being able to use the advice's return value for assignment is that if there's only one assignment, there's not even a chance for JIT to miss the EA. As you said, the instrumented methods have to be sufficiently short, for example.

`@Enter` would still work btw, but the exit advice might have to accept an `Object[]`.

If it's not too complex, what about using both options? I.e. supporting `@Advice.AssignTo` on the return value and an `Object[]` advice method argument.

---

Another approach is trying to almost transparently support `readOnly` for non-inlined advices.

Given an advice:
```java
@Advice.OnMethodEnter(inline = false)
public static void onEnter(
  @Advice.Argument(value = 0, readOnly = false) String foo, 
  @Advice.Argument(value = 1, readOnly = false) String bar,
  @Advice.Binder Object[] binder) {
    foo = "foo";
    bar = "bar";
}
```

When applied to a method:
```java
public String concat(String foo, String bar) {
    return foo + bar;
}
```

Could we turn it into:
```java
public String concat(String foo, String bar) {
    Object marker = new Object();
    Object[] binder = new Object[] {marker, marker};
    onEnter(foo, bar, binder);
    if (binder[0] != marker) foo = binder[0];
    if (binder[1] != marker) bar = binder[1];
    return foo + bar;
}
```

And re-write the advice method to
```java
@Advice.OnMethodEnter(inline = false)
public static void onEnter(
  @Advice.Argument(value = 0) String foo, 
  @Advice.Argument(value = 1) String bar,
  @Advice.Binder Object[] binder) {
    binder[0] = "foo";
    binder[1] = "bar";
}
```

---

What about turning `Object marker = new Object();` into a constant?

PR #1120, opened 2021-09-27
CI simplifications

scordio:
I recently took inspiration from this repo to create a workflow based on the `maven-release-plugin` and I thought I would share a few simplifications that caught my attention:
* The GPG private key import can be delegated to `actions/setup-java` via the [`gpg-private-key` input parameter](https://github.com/actions/setup-java/blob/main/docs/advanced-usage.md#gpg)
* The [caching feature](https://github.com/actions/setup-java#caching-packages-dependencies) of `actions/setup-java` can be used instead of the [default `actions/cache` configuration](https://github.com/actions/cache/blob/main/examples.md#java---maven)

There is one more `actions/cache` configuration that I didn't change since it has a non-default key pattern:

https://github.com/raphw/byte-buddy/blob/90afb6039d420a43682dbc3055765355b87eb094/.github/workflows/main.yml#L107-L111

Not sure if a non-default pattern was chosen on purpose. If not, that could be simplified as well.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -16,16 +16,12 @@ jobs:
     runs-on: ubuntu-18.04
     steps:
       - uses: actions/checkout@v2
-      - uses: actions/cache@v2
-        with:
-          path: ~/.m2/repository
-          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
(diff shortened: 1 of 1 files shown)


raphw:
Brilliant. The non-default pattern is used to avoid that libraries that are downloaded via HTTP leak into, for example, the release build.

PR #1181, opened 2021-12-27
Also check maven.compiler.release when looking for java version

famod:
https://maven.apache.org/plugins/maven-compiler-plugin/compile-mojo.html#release
https://www.baeldung.com/java-compiler-release-option

For most (Java 9+) projects, the `release` property is the way to go these days, IMO.


raphw:
Good catch, I was under the wrong impression that the target property is set implicitly when using the release property, but this might not be the case. Can both properties be set and how would this affect the compilation target?

famod:
They can bot be set, but as soon as `release` is set, only that flag will be passed to the compiler. This can be verified via `mvn -X ...` (look for `[DEBUG] Command line options:`).
So `maven-compiler-plugin` basically ignores `target` when `release` is set (which is not entirely true, see https://issues.apache.org/jira/browse/MCOMPILER-413, but that's not relevant here).

PR #1279, opened 2022-07-06
Add CodeSizeEvaluator to remapped ASM commons

geoand:
When working on a ByteBuddy plugin (for Quarkus) to be used with the bytebuddy-maven-plugin, 
I encountered the need to use the `org.objectweb.asm.commons.CodeSizeEvaluator`, which has
not been included in the shaded jars.
This PR makes sure it's included

### DIFF
--- a/byte-buddy/pom.xml
+++ b/byte-buddy/pom.xml
@@ -276,6 +276,7 @@
                                             <includes>
                                                 <include>org/objectweb/asm/commons/AnnotationRemapper.**</include>
                                                 <include>org/objectweb/asm/commons/ClassRemapper.**</include>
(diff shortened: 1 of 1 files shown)


raphw:
I did not add the entire commons to Byte Buddy on purpose, only the renaming bits which Byte Buddy requires internally.

Could you simply copy the class to your plugin (with reference)? It's only 200 lines and I wanted to keep the public ASM surface rather small, especially with the OpenJDK API on the horizon.

geoand:
Hi, 

Yes, I completely understand if you want to close the PR. 
I have indeed copied the code so I am not blocked in any way.

raphw:
Thanks, I think that's the best approach!

geoand:
👍🏼

PR #1290, opened 2022-07-27
Add android support to gradle plugin

LikeTheSalad:


raphw:
Woop, thanks for that. Unfortunately, I reworked the Gradle plugin and build a bit just the last days. I hope that it's an easy adjustment. Could you also squash the commits, that would make it easier to review in my IDE.

A few first comments:
 - Is it possible to avoid Kotlin in the project? One less thing to worry about.
 - We can let the build fail for now, no need to javadoc all the things yet.
 - It still looks to me that there is two plugins now, the Android Byte Buddy plugin and the "regular" one. Ideally, I would want to avoid this and make the regular plugin detect if it should run in Android mode. Do you think that's a possibility?

I am still partially on vacation this week but I will have a good look some time soon!

LikeTheSalad:
Hi @raphw, yeah no worries, your recent changes didn't cause much trouble because I haven't yet made the connection between the new plugin and the main one :) it's actually one of the things I'd like to discuss with you first before making those changes, that's why they are not connected yet, but by the time this PR is merged, the outcome should be a single plugin that detects when it should go Android mode.

Regarding Kotlin, yeah it is possible to avoid using Kotlin, though I'm not going to lie, it's kind of annoying having to write Java after having made the switch to mostly Kotlin for a while now 😅, but I see your point, I think I can make that change in the meantime.

Since you're partially on vacation, please let me know once you're available to discuss the final details 👍

LikeTheSalad:
Hi @raphw, I've just uploaded changes for removing Kotlin and squashing all the commits, I'm planning tomorrow to push some changes regarding connecting the android plugin to the main plugin so that we can have an idea of how it could be done.

raphw:
You already inspired a bit of refactoring. The class loader scanning is now done from the core lib.

LikeTheSalad:
Cool! I've made changes to use the new method to find plugins 👍 I've also connected the android plugin to the main one, I left the recent commits unsquashed in case you wanted to take a look at them to only see my recent changes, though I can squash those too if needed.

raphw:
Continuing the work on this branch: https://github.com/raphw/byte-buddy/compare/android-plugin

PR #1295, opened 2022-08-02
update usage documentation

a364176773:


a364176773:
@raphw PTAL , I found that the document was a little behind, so I updated it

raphw:
You are right, it is. Thanks for updating. I was still wondering why you needed this additional exception block with the rethrow.

a364176773:
> You are right, it is. Thanks for updating. I was still wondering why you needed this additional exception block with the rethrow.

Thanks for your review. I think that after describing it in this way, the user can clearly know that this exception can be handled extra, rather than thrown out directly

raphw:
Would you remove the changes on the intereptor? Then I can merge this.

a364176773:
> Would you remove the changes on the intereptor? Then I can merge this.

done

raphw (approved):

PR #1296, opened 2022-08-02
update example, fix #1293

### LINKED ISSUE #1293: The example in "Working with unloaded classes" is wrong.
The example in "Working with unloaded classes" is wrong.

```java
class MyApplication {
  public static void main(String[] args) {
    TypePool typePool = TypePool.Default.ofSystemLoader();
    Class bar = new ByteBuddy()
      .redefine(typePool.describe("foo.Bar").resolve(), // do not use 'Bar.class'
                ClassFileLocator.ForClassLoader.ofSystemLoader())
      .defineField("qux", String.class) // we learn more about defining fields later
      .make()
      .load(ClassLoader.getSystemClassLoader());
    assertThat(bar.getDeclaredField("qux"), notNullValue());
  }
}
```

diguage:


raphw:
I could run the sample without issues.

Issue #1347, opened 2022-10-24
Class initialization deadlock with TypeDescription.ForLoadedType and TypeDescription.Generic


cushon:
We started seeing infrequent deadlocks in code using bytebuddy after upgrading to a version of bytebuddy built targeting Java 8+ bytecode.

My colleague @jianglizhou figured out that this is do to a circular class initialization deadlock between `net.bytebuddy.description.type.TypeDescription$ForLoadedType` and `net.bytebuddy.description.type.TypeDescription$Generic` that manifests when there are two calls to `net.bytebuddy.ByteBuddy.subclass`, one for an interface, and one for a non-interface class:

https://github.com/raphw/byte-buddy/blob/eea667f8c5e3a95512119c585e233ab10b33f6e4/byte-buddy-dep/src/main/java/net/bytebuddy/ByteBuddy.java#L511-L516

This is a sample thread dump showing the deadlock:

```
"pool-1-thread-1@1152" prio=5 tid=0x11 nid=NA runnable
  java.lang.Thread.State: RUNNABLE
      at net.bytebuddy.description.type.TypeDescription$ForLoadedType.asGenericType(TypeDescription.java:8913)
      at net.bytebuddy.ByteBuddy.subclass(ByteBuddy.java:515)
      at net.bytebuddy.ByteBuddy.subclass(ByteBuddy.java:483)
      at net.bytebuddy.ByteBuddy.subclass(ByteBuddy.java:380)
      at net.bytebuddy.ByteBuddySubclassDeadlockTest$Tester$1.call(ByteBuddySubclassDeadlockTest.java:38)
      at net.bytebuddy.ByteBuddySubclassDeadlockTest$Tester$1.call(ByteBuddySubclassDeadlockTest.java:34)
      at java.util.concurrent.FutureTask.run$$$capture(FutureTask.java:264)
      at java.util.concurrent.FutureTask.run(FutureTask.java:-1)
      at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
      at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
      at java.lang.Thread.run(Thread.java:833)

"pool-1-thread-2@1157" prio=5 tid=0x12 nid=NA runnable
  java.lang.Thread.State: RUNNABLE
      at net.bytebuddy.description.type.TypeDescription$Generic.<clinit>(TypeDescription.java:476)
[...]

PR #1348, opened 2022-10-24
Break a class initialization deadlock

cushon:
https://github.com/raphw/byte-buddy/issues/1347

### DIFF
--- a/byte-buddy-dep/src/main/java/net/bytebuddy/ByteBuddy.java
+++ b/byte-buddy-dep/src/main/java/net/bytebuddy/ByteBuddy.java
@@ -146,6 +146,9 @@ public class ByteBuddy {
     @MaybeNull
     private static final Implementation.Context.Factory DEFAULT_IMPLEMENTATION_CONTEXT_FACTORY;
 
+    // break a class initialization cycle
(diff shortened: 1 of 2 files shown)


raphw:
I would prefer a fix as suggested in https://github.com/raphw/byte-buddy/commit/5a809653d383454a2f196fc0366b2721c779871c

cushon:
> I would prefer a fix as suggested in [5a80965](https://github.com/raphw/byte-buddy/commit/5a809653d383454a2f196fc0366b2721c779871c)

Okay, thanks, I see that's been committed.

Do you want the test, or should I close the PR?

raphw:
I tried to solve this differently, thanks!

cushon:
> > Do you want the test, or should I close the PR?
>
> I tried to solve this differently, thanks!

Ok, thanks.

It looks like the test did get added in https://github.com/raphw/byte-buddy/commit/50acab840e4d3e241c03f484dcad9b0d0f923dbf.

raphw:
Yes, I'm a bit uncertain if I want to keep it as it's flaky by definition, but until changed my mind to keep it until it's getting in the way at least.

PR #1371, opened 2022-12-16
upgrade maven-release-plugin

hboutemy:
this will update the Reproducible Builds outputTimestamp value during releases: this should fix issue from https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/net/bytebuddy/README.md


raphw:
I'd prefer to mot merge a milestone release. But it cannot be long to an actual one.

hboutemy:
honestly, I don't know when we'lll do the final 3.0.0, what i can tell you is that 3.0.0-M7 is really worth it :)

PR #1389, opened 2023-01-10
Providing Android runtimeClasspath using new tools from API 7.3

LikeTheSalad:


raphw:
I changed it by relying on the reflection API being able to resolve a method. Does this still work as intended?

LikeTheSalad:
That should work too. I think what should be avoided the most would be casting the Variant object to ComponentImpl, but apart from that, the way to check for the new method should be fine either way

raphw:
I cast to the interface now that declares the method but also added a type check to give a better error message. 

As long as the new method is available, it will be used. I'll make a release this week!

PR #1391, opened 2023-01-12
Add public addURL to BOOT_LOADER_PROXY

aarolesl:
We own an Amazon open-source project which utilizes ByteBuddy, and this PR for ByteBuddy solves an obstacle we have encountered in supporting Java 17. Our project: [https://github.com/awslabs/disco](https://github.com/awslabs/disco)

Disco is an open source event-driven toolkit. Disco provides a framework for plugins which may use features offered by the toolkit, the primary of which being Context Propagation (across threads and services) and Events. Disco vends a Java Agent which services use, where plugins are loaded by the Disco Java Agent and may themselves be thought of as agents. Installables - components of some plugins - utilize ByteBuddy to perform instrumentations (for example, a plugin may instrument Apache to propagate context through service calls.) Installable Interface: [https://github.com/awslabs/disco/blob/master/disco-java-agent/disco-java-agent-plugin-api/src/main/java/software/amazon/disco/agent/interception/Installable.java#L35](https://github.com/awslabs/disco/blob/master/disco-java-agent/disco-java-agent-plugin-api/src/main/java/software/amazon/disco/agent/interception/Installable.java#L35)

Some plugins have a requirement to exist on the bootstrap classloader and may be injected to it at runtime during the PluginDiscovery phase of the Disco Java Agent. [https://github.com/awslabs/disco/blob/master/disco-java-agent/disco-java-agent-core/src/main/java/software/amazon/disco/agent/plugin/PluginDiscovery.java#L311](https://github.com/awslabs/disco/blob/master/disco-java-agent/disco-java-agent-core/src/main/java/software/amazon/disco/agent/plugin/PluginDiscovery.java#L311)

Since these plugins can define installables which utilize ByteBuddy to perform instrumentations, they commonly define classes which need to be loaded as resources by ByteBuddy (e.g. Advices). Because the bootstrap classloader does not support resource loading, just classloading, we rely on ByteBuddy’s BOOT_LOADER_PROXY. When bootstrap plugins are loaded during runtime, we reflectively call addURL on ByteBuddy’s BOOT_LOADER_PROXY, since ByteBuddy will check here by-default for performing bootstrap resource loading. [https://github.com/awslabs/disco/blob/master/disco-java-agent/disco-java-agent-inject-api/src/main/java/software/amazon/disco/agent/inject/Injector.java#L144](https://github.com/awslabs/disco/blob/master/disco-java-agent/disco-java-agent-inject-api/src/main/java/software/amazon/disco/agent/inject/Injector.java#L144)

This saves plugin developers the undifferentiated effort of utilizing a ClassFileLocator in their installables and needing to be aware of what mechanism the Disco Java Agent is using to load them. By performing this work when loading the plugins, they can be injected onto the bootstrap classloader at runtime by Disco Java Agent, and interface with the AgentBuilders that we vend to them in the installable interface without themselves carrying specific logic around how they were loaded.

However, with JDK 17, this reflective access is prevented, as addURL is protected in URLClassLoader and access modifiers are more strongly enforced. The change here would enable us to continue to ensure that these bootstrap plugins can be resource-loaded with the AgentBuilders we supply to them by-default without needing all plugin developers to take on work to update their usages of ByteBuddy to use a ClassFileLocator which matches how we have loaded their plugins.

### DIFF
--- a/byte-buddy-dep/src/main/java/net/bytebuddy/dynamic/ClassFileLocator.java
+++ b/byte-buddy-dep/src/main/java/net/bytebuddy/dynamic/ClassFileLocator.java
@@ -476,7 +476,29 @@ public interface ClassFileLocator extends Closeable {
              * {@inheritDoc}
              */
             public ClassLoader run() {
-                return new URLClassLoader(new URL[0], ClassLoadingStrategy.BOOTSTRAP_LOADER);
+                return new BootLoaderProxy(new URL[0], ClassLoadingStrategy.BOOTSTRAP_LOADER);
+            }
+
+            /**
+             * A URLClassLoader which exposes addURL to be used as a Bootloader Proxy.
+             */
+            class BootLoaderProxy extends URLClassLoader {
+                /**
+                  * Constructs a new BootLoaderProxy for the given URLs and parent ClassLoader.
+                  * @param      urls the URLs from which to load classes and resources
+                  * @param      parent the parent class loader for delegation
+                  */
+                BootLoaderProxy(URL[] urls, ClassLoader parent) {
+                    super(urls, parent);
+                }
+
+                /**
+                 * {@inheritDoc}
+                 */
+                @Override
+                public void addURL(URL url) {
+                    super.addURL(url);
+                }
             }
         }
 


raphw:
I am not sure if I wanted to introduce mutability here. But `ClassLoadingDelegate` is a public interface. Would it be an option for you to implement it yourself?

aarolesl:
This is understandably not an area where mutability had been intended and not a directly supported use case. I am not sure that implementing `ClassLoadingDelegate` would be helpful for our use case, as we equally have the option of defining a `ClassFileLocator` which has visibility into the correct resources. 

However, today we allow plugins (acting like mini java agents) for Disco to interact quite generally with `AgentBuilder`s that we supply to them, and we allow them to expect that resources which they define can be found on the bootstrap classloader by ByteBuddy by default. Supplying to these plugins a `ClassFileLocator` or `ClassLoadingDelegate`  which has visibility into their resources which they must use in their interactions with ByteBuddy breaks some of the abstraction we try to maintain over how we load their plugins and asks many Disco plugins to change how they interact with the supplied AgentBuilders.

The requested changes would indeed still require us to use reflection, but by making `addURL` public on this specific BOOT_LOADER_PROXY instance in `ClassFileLocator.ForClassLoader`, none of these reflection operations are blocked by JDK 17. In essence, this is a hack-around to mimic (from the perspective of a Disco plugin) being able to inject the plugin Jar to the bootstrap classloader directly.

raphw:
I understand. I still wonder if there's a better solution.

So if I understand you correctly, you want to receive an agent builder, and then change that builder to also resolve additional class files from the boot loader that you injected?

aarolesl:
We want the agent builder to be passed to plugins - which we do not have control over how they use the agent builder exactly - to be able to resolve class files that we injected onto the boot loader, yes. Because these plugins can do a wide variety of actions with the agent builder that they receive, we did not find a mechanism which reliably "primes" the agent builder to locate the boot-loader-injected jars other than adding them to the BOOT_LOADER_PROXY. Supplying a LocationStrategy to the agent builder before passing it to the plugins, for example, still appears to allow plugins to interact with the agent builder in some ways that do not cause it to resolve the injected resources.

raphw:
I see. I just added a `with(ClassFileLocator)` method to the `AgentBuilder` API which allows to add types which are not otherwise visible. I find that a cleaner approach than just modifying the boot loader proxy.

The change is pushed to master. Could it work for you?

aarolesl:
That looks like a great change, and gets closer to allowing us to prime the `AgentBuilder` we supply to the plugins. However, for our use case sometimes plugins take actions which attempt to resource-load class files in a way that still will not end up using the `ClassFileLocator` that was supplied to the AgentBuilder. One fairly common example is:
```
public AgentBuilder install(AgentBuilder agentBuilder) {
        return agentBuilder
            .type(ElementMatchers...)
            .transform((builder, typeDescription, classLoader, module) -> builder
                .method(ElementMatchers...)
                .intercept(Advice.to(SomeClass.class))
            )
```
Here, even if we add `SomeClass` to our own equivalent classloader to the boot loader proxy (say `DiscoBootstrapParentedClassLoader`) for resource loading after injecting it to the bootstrap classloader, and we then supply that to the AgentBuilder that is passed to the plugin via a ClassFileLocator, Advice.to will see that `SomeClass` is on the bootstrap classloader and fail to resource load it since it is only visible for resource-loading via the `DiscoBootstrapParentedClassLoader` and Advice.to will instead use the `BOOT_LOADER_PROXY`. 

I am not sure that there is a nice mechanism by which to ensure that the right `ClassFileLocator` ends up being used for these types of use cases, other than asking every plugin to change their code to additionally supply it at every point of interception, since other APIs outside of `AgentBuilder` may trigger the resource-loading without using the `ClassFileLocator` of the surrounding `AgentBuilder`. Hence why today we centrally target the `BOOT_LOADER_PROXY` to ensure each of these types of use cases are caught.

raphw:
When using `Advice.to(SomeClass.class)`, the boot loader proxy is used but that is rather a coincidence. If the advice class is not on the boot loader, it would not use that proxy and miss the classes. The only reliable approach to this would be to supply a custom `ClassFileLocator` as a second argument to the `Advice`.

Generally, I recommend using `AgentBuilder.Transformer.ForAdvice` in this case where the resolution is more open to modifications.

aarolesl:
It sounds like then the supported approach would be us placing stronger restrictions on the interactions plugins make with (e.g.) `Advice` or for us to supply the appropriate `ClassFileLocator` to the plugins to use if they absolutely need. We'll explore options in that area then looking forward. The changes to `AgentBuilder` to support pre-defining a `ClassFileLocator` should be helpful in meeting our use case in the middle, we appreciate those changes!

I'll go ahead and close this PR since it sounds like directly modifying `BOOT_LOADER_PROXY` is not the preferred route here for ByteBuddy.

raphw:
Thanks, I think this is a better approach. Otherwise, we would introduce some form of global state what can have rather unexpected consequences if the implementation of `Advice` would change, for example. Or if a user of your's would use their own `ClassFileLocator`.

I think ideally your users are aware of this and adjust their code to make this dependency explicit.

Issue #1418, opened 2023-03-09
Is there an API to get all instrumented methods?


FrankChen021:
Hi @raphw 

By listening on `AgentBuilder.Listener#onTransformation`, we can know if the clazz has been instrusmented. But if I want to know which methods have been instrumented in this class, is there any API in byte-buddy to do so?

chatgpt suggests me to add a customer annotation to the target method during transformation, and then in the onTransformation method, we can check if the methods on the given type have been annotated by that customer annotation. I think it's less efficient, and want to know if there's a better way to do so?

FrankChen021:
Thank you @raphw for the response. 

I don't know where did the chatGPT get the idea, and solution it suggested is really a little bit wired.

By looking through the byte-buddy API, I came across this solution.

before

```java
                    builder = builder.visit(Advice.withCustomMapping()
                                                  .to(MethodDecoratorAdvice.class)
                                                  .on(matcher);

```

after

```java
                    Advice advice = Advice.withCustomMapping()
                                          .to(MethodDecoratorAdvice.class);
                    builder = builder.visit(new AsmVisitorWrapper.ForDeclaredMethods().invokable(matcher,
                                                                                                 advice, 
                                                                                                 new Visitor()));

        static class Visitor implements AsmVisitorWrapper.ForDeclaredMethods.MethodVisitorWrapper {

            @Override
            public MethodVisitor wrap(TypeDescription instrumentedType,
                                      MethodDescription instrumentedMethod,
                                      MethodVisitor methodVisitor,
                                      Implementation.Context implementationContext,
                                      TypePool typePool,
                                      int writerFlags,
                                      int readerFlags) {

                // here the instrumentedMethod is what I want to get
                return methodVisitor;
            }
        }

```

raphw:
Did you consider remembering all this in a map?

FrankChen021:
> Did you consider remembering all this in a map?

Yes, above code is just an illustration.

FrankChen021:
Hi @raphw , I'm adding this to my project, and there's a problem.

In my Advice, there're some customized annotations and resolvers, for example

```java
public class AfterAdvice {

    @Advice.OnMethodEnter
    public static void onEnter(
        @AdviceAnnotation.Name String name,
        @Advice.This(optional = true) Object target
    ) {
        ....
    }
}
```

and the corresponding resolver for above annotation is defined as 

```java
  public static class NameResolver implements Advice.OffsetMapping {
      private final String name;

[...]

PR #1427, opened 2023-04-09
Allow to access offset mappings in custom `MethodVisitorWrapper`

FrankChen021:
This PR exposes some inner classes in Advice, and passes `offsetMapping` explicitly to `AdviceMethodInliner` so that user defined visitor that inhertis `MethodVisitorWrapper` can access offset mappings to extract informations from custom Advice annotations. See #1418

Two test cases are added to demonstrate the use case. In the test case, there's a visitor that calls the exposed classes/methods.

This is the mininal work to make it, may not be the best ways.


raphw:
As explained in the related issue, I do not want to expose this API.

PR #1475, opened 2023-07-08
fixing the flaky tests that occured due to DeclaredFields

swethakare:
getDeclaredFields does not ensure the order of the fields that are fetched during the execution of junit.
performing a sort on the output of the getDeclaredFields then performing an assertion prevents this non-deterministic error.


swethakare:
Wrong repo pr

PR #1484, opened 2023-07-14
Fixed failure to generate get methods for fields with capitalized initial letters

haiyanghan:
Fixed failure to generate get methods for fields with capitalized initial letters

### DIFF
--- a/byte-buddy-dep/src/main/java/net/bytebuddy/implementation/FieldAccessor.java
+++ b/byte-buddy-dep/src/main/java/net/bytebuddy/implementation/FieldAccessor.java
@@ -19,6 +19,7 @@ import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import net.bytebuddy.build.HashCodeAndEqualsPlugin;
 import net.bytebuddy.description.field.FieldDescription;
 import net.bytebuddy.description.method.MethodDescription;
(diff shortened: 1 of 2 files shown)


haiyanghan:
@raphw hi, can this pr be merged?

raphw:
I did not have time yet, but was planning to look into this early August. Why did you close this ticket?

haiyanghan:
> 

Sorry, I thought this pr would not be merged

raphw:
I will look into this, but the suggested solution is too simple. Fields can be declared by super types, and the traversal would be duplicated, what I want to avoid. I do not see a trivial solution, though.

haiyanghan:
> I will look into this, but the suggested solution is too simple. Fields can be declared by super types, and the traversal would be duplicated, what I want to avoid. I do not see a trivial solution, though.

I'm not familiar with the source code of bytebuddy, so used a relatively simple wording

haiyanghan:
> I will look into this, but the suggested solution is too simple. Fields can be declared by super types, and the traversal would be duplicated, what I want to avoid. I do not see a trivial solution, though.


I see you have resubmitted this issue, I have closed this PR

raphw:
I have solved it slightly different, but thanks for your suggestion and finding this corner-case.

I am actually not sure if fields with a first capital letter a Java bean compliant, but I think it is worth supporting.

PR #1510, opened 2023-08-22
Fix reading bytecode version for java1

SylvainJuge:
The Java class file format ([doc](https://jcp.org/aboutJava/communityprocess/maintenance/jsr924/JVMS-SE5.0-Ch4-ClassFile.pdf)) stores the bytecode version in the bytes 4 to 8 just after the famous `0xCAFEBABE` magic value.
In practice, the minor version hasn't been used since Java1, and both minor and major can be read as single bytes.

When reading the version of an existing class compiled for Java 1, the minor version (3) is read at the wrong offset and is thus ignored, however the major version (45) is properly read, which makes it as if we were attempting to read a class that was compiled for Java 1.0 ([doc](https://javaalmanac.io/bytecode/versions/)), which is not covered by ASM as there isn't any `Opcodes.V1_0`.

This PR fixes the bug in version reading and adds a test to cover it.

### DIFF
--- a/byte-buddy-dep/src/main/java/net/bytebuddy/ClassFileVersion.java
+++ b/byte-buddy-dep/src/main/java/net/bytebuddy/ClassFileVersion.java
@@ -402,7 +402,7 @@ public class ClassFileVersion implements Comparable<ClassFileVersion>, Serializa
         if (binaryRepresentation.length < 7) {
             throw new IllegalArgumentException("Supplied byte array is too short to be a class file with " + binaryRepresentation.length + " byte");
         }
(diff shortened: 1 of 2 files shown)


raphw:
Good catch! I just moved the test to the parameterized version.

SylvainJuge:
Thanks for merging this !

This issue is linked to the 1.14.6 milestone, however this one has been released last week (and it's not included in the release notes), does it means it should actually be part of the 1.14.7 (or whatever comes next) release ?

raphw:
I use this feature poorly where I mark the version where this was reported as I normally solve the issue within the next release. Pushing a release now.

PR #1566, opened 2023-11-27
add unit test for System class redefinition

sullis:
This unit test validates that the Byte Buddy Agent can
redefine java.lang.System


raphw:
Unfortunately, this might interfere with others tests.This is why I prefer the testing on class loader isolating level.

PR #1578, opened 2024-01-02
Fix typo in Javadocs

perlun:
Thanks for a nice tool. :+1: When looking at these docs (trying to get my code working :sweat_smile:), I saw this. Given that the `SuperCall` entry right below refers to `DefaultCall`, I guess this one should have been `DefaultCall` (instead of two `SuperCall` entries)?


raphw:
the javadoc was correct actually, but there was a confusion with the subsequent section. SuperCall can indeed invoke non-ambiguous default methods.

PR #1613, opened 2024-04-03
Update InstrumentedType.java to check instrumented classfile is in valid Unicode namespace instead

leerjae:
We own an Amazon open-source project which utilizes ByteBuddy, and this PR for ByteBuddy solves an obstacle we have encountered while supporting our customers Our project: https://github.com/awslabs/disco

In some cases we need to instrument shaded classes with FQN such as `*.!internal.*` that are [valid JVM classfile names](https://docs.oracle.com/javase/specs/jvms/se7/html/jvms-4.html#:~:text=from%20the%20entire-,Unicode%20codespace,-.%20Class%20and%20interface) but are not accepted by the compiler. This is intentionally done to prevent external dependencies from using the shaded classes. These classes’ bytecode are valid and are processed and have its bytecode changed by ASM but failing `InstrumentedType.validate()` post-instrumentation validation because of i`sValidIdentifier()` which only allows Java Identifier characters. Is it possible to add slack to the instrumented classfile name validation to allow characters from the Unicode namespace instead? According to the Oracle's JVM specification, classfile names can be drawn from all of the Unicode namespace (see link above).

### DIFF
--- a/byte-buddy-dep/src/main/java/net/bytebuddy/dynamic/scaffold/InstrumentedType.java
+++ b/byte-buddy-dep/src/main/java/net/bytebuddy/dynamic/scaffold/InstrumentedType.java
@@ -1842,13 +1842,13 @@ public interface InstrumentedType extends TypeDescription {
          * @return {@code true} if the given identifier is valid.
          */
         private static boolean isValidIdentifier(String identifier) {
-            if (KEYWORDS.contains(identifier) || identifier.length() == 0 || !Character.isJavaIdentifierStart(identifier.charAt(0))) {
+            if (KEYWORDS.contains(identifier) || identifier.length() == 0 || !Character.isUnicodeIdentifierStart(identifier.charAt(0))) {
                 return false;
             } else if (identifier.equals(PackageDescription.PACKAGE_CLASS_NAME)) {
                 return true;
             }
             for (int index = 1; index < identifier.length(); index++) {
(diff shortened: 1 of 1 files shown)


raphw:
Makes sense to me as Byte Buddy follows the class file specification rather than the language specification. I was not aware of this.

leerjae:
Hi raphw, thank you for quickly accomodating the pull request. Also I've been trying to understand the [additional type validation that BB does](https://javadoc.io/static/net.bytebuddy/byte-buddy/1.6.5/net/bytebuddy/dynamic/scaffold/TypeValidation.html#:~:text=Byte%20Buddy%27s%20checks%20might%20be%20more%20expressive%20in%20the%20context%20of%20using%20the%20library) outside of what the JVM provides.  I've been digging into the [JVM specs](https://docs.oracle.com/javase/specs/jvms/se7/html/jvms-4.html#jvms-4.10) and InstrumentType.validate() source code to try and understand how BB type validation is more expressive but having hard time wrapping my head around it. At a high level can you tell me how "Byte Buddy's checks might be more expressive in the context of using the library"?

raphw:
The validation will create a "better" error message than the `VerificationError` that the JVM emits. This is less true in newer JVM versions where the error message by the JVM is fairly understandable.

The main difference however is that a verification error in the JVM will break the class forever. If Byte Buddy fails verification during an agent's application for example, the instrumentation is aborted instead.

PR #1616, opened 2024-04-19
Update NamedPipe handling to allow callback to Service Accounts

FelixMarxIBM:
While testing attachment via byte-buddy on Windows, we noticed that the NamedPipe is not able to report back the current state if the attachment was started from a service account.
This is possible with the normal attach-tools.jar so I investigated what the difference is and it turns out that the hotspot attachment uses named pipes with a SecurityDescriptor.
See [VirtualMachineImpl.c](https://github.com/AdoptOpenJDK/openjdk-jdk11/blob/master/src/jdk.attach/windows/native/libattach/VirtualMachineImpl.c#L269-L281) how that is done.
This PR does the same thing in JNA and in our test the attachment works afterwards when started from a service account against a JVM started from a user.

`VirtualMachineImpl.c`:
```
    // Custom Security Descriptor is required here to "get" Medium Integrity Level.
    // In order to allow Medium Integrity Level clients to open
    // and use a NamedPipe created by an High Integrity Level process.
    TCHAR *szSD = TEXT("D:")                  // Discretionary ACL
                  TEXT("(A;OICI;GRGW;;;WD)")  // Allow read/write to Everybody
                  TEXT("(A;OICI;GA;;;SY)")    // Allow full control to System
                  TEXT("(A;OICI;GA;;;BA)");   // Allow full control to Administrators
```

### DIFF
--- a/byte-buddy-agent/src/main/java/net/bytebuddy/agent/VirtualMachine.java
+++ b/byte-buddy-agent/src/main/java/net/bytebuddy/agent/VirtualMachine.java
@@ -1007,6 +1007,9 @@ public interface VirtualMachine {
                         throw new IllegalArgumentException("Cannot supply more then four arguments to Windows attach mechanism: " + Arrays.asList(argument));
                     }
                     String name = "\\\\.\\pipe\\javatool" + Math.abs(random.nextInt() + 1);
+
+                    WinBase.SECURITY_ATTRIBUTES sa = createSecurityAttributesToAllowMediumIntegrity();
+
                     WinNT.HANDLE pipe = Kernel32.INSTANCE.CreateNamedPipe(name,
                             WinBase.PIPE_ACCESS_INBOUND,
                             WinBase.PIPE_TYPE_BYTE | WinBase.PIPE_READMODE_BYTE | WinBase.PIPE_WAIT,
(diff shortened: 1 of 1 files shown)


FelixMarxIBM:
@raphw should I create an Issue for that or is a PR enough?

raphw:
Thanks, I come across windows so rarely, great you could contribute this! I'll merge it obce the checks complete.

FelixMarxIBM:
Hi @raphw I saw that the build failed [here](https://github.com/raphw/byte-buddy/actions/runs/8766420402/job/24058539548)
with the error
`Error:  /home/runner/work/byte-buddy/byte-buddy/byte-buddy-agent/src/main/java/net/bytebuddy/agent/VirtualMachine.java:1083: @return tag should be present and have description. [JavadocMethod]
`
I'll quickly create a follow up PR to fix that.

FelixMarxIBM:
Hi @raphw I think the milestone is wrong here and should be `1.14.14`, not `1.12.14`

raphw:
Do not use `byte-buddy-dep` but `byte-buddy`. In your code, use `ClassVisitorFactory` to bridge between different shadings of ASM.

PR #1677, opened 2024-07-16
Fix bug in Android SDK > 28

Dev4Mod:
This flag is for dex with permission to access restricted APIs, however this causes an error on Android 9 or higher. It was changed because it was not necessary


raphw:
I hope this has no side effects, but I do not have an easy way of verifying this, so I rely on your assertion where users can roll back the version of this is a problem. So we just try and see the reception.

PR #1788, opened 2025-03-20
Comprehensive documentation update on gh-pages website

manoelcampos:
Hello, I'm re-reading Byte Buddy's documentation and found some issues I can fix.
I'm sending it in advance in order to discuss some issues I've found.
You can check the live changes at https://manoelcampos.github.io/byte-buddy/

# Some issues to discuss

## JDK version on GitHub Workflow

If you take a look at https://manoelcampos.github.io/byte-buddy/#/develop, it currently states that:

> The GitHub Action is configured to compile and test the project for the Open JDK versions 6 and 7 and the Oracle JDK versions 7 and 8.

This information is not accurate, since there are tests for the Oracle JDK 8, 11, 17, 21 (hotspot-supported version).
But there are tests for JDK 6 and 7 (hotspot-legacy). In fact, this workflow provides lots of tests for different JDK distributions and versions. Maybe, the documentation would just state that:

> The GitHub Action is configured to compile and test the project for different versions of Open JDK and the Oracle JDK.

- [x] The sentence was changed to: "The GitHub Action is configured to compile and test the project for different versions of OpenJDK, Oracle JDK and other distributions, starting from JDK 6 up to the latest release (despite deprecated JDK releases are not supported anymore)."

## Minimum JDK version required

The welcome page at https://manoelcampos.github.io/byte-buddy/ says that:

> Byte Buddy is written in Java 5

Contradictorily, https://manoelcampos.github.io/byte-buddy/#/develop, states:

> All this implies that you need to write Java code that is compatible to a Java 6 compiler.

- [x]  Adds variable inside app.js to indicate ByteBuddy is a Java 5 lib.

## Develop page

- [x] I'm not sure what you meant here: "All code is to be documented what makes this property [easier to validate](https://en.wikipedia.org/wiki/Broken_windows_theory) by automatic checks"
- [x] "Java 9 support is currently still experimental." Sentence removed
- [x] Is the following sentence still true?

> As of Java 8, Byte Buddy not currently support the following functionality.

## Tutorial page

The [link about inflation mechanism used for reflection](https://manoelcampos.github.io/byte-buddy/#/tutorial) is broken.
The only reference I could find was [this one](https://www.ibm.com/docs/en/ztpf/2023?topic=options-dsunreflectinflationthreshold).

However, this inflation mechanism seems to be dropped in [JDK 18](https://openjdk.org/jeps/416), being replaced by method handles (I'm not sure about that). This way, I think the sentence below needs to be rewritten:

> However, modern JVMs know a concept called inflation where the JNI-based method invocation is replaced by generated byte code that is injected into a dynamically created class.

- [x] I updated the entire paragraph including information about JDK releases before and after 18.

- [x] It's not clear what you mean by "Benchmark (2a)" and other references. There are no such references on the table.
- [x] The sentence "Also, if example.Bar extended example.Foo, any overridden package-private methods would become inoperative but would delegate to the original implementations." is confusing.
- [x] Is the following sentence still true? "Furthermore, Byte Buddy currently only takes loaded classes as its argument, this will change in future versions where the existing API can be used to equally work with unloaded classes."
- [x] At "From the above example, it is obvious that the super method is called by injecting some instance of a Callable into the LoggerInterceptor which invokes the original, non-overridden implementation of MemoryDatabase#load(String) from its call method. This helper class is called an AuxiliaryType within Byte Buddy's terminology", it's not clear which AuxiliaryType is that. Is it a class that ByteBuddy creates that implements Callable?
- [x] About the sentence "You might want to filter the range of possible candidate methods by calling filter(ElementMatcher) on any MethodDelegation". There is not `filter` method on `MethodDelegation` anymore. I couldn't find a different way to do that. Is this feature still supported somehow?

- [x] Is the sentence below still true? Since at the develop page you say that the lib is "feature-complete".

> Furthermore, we are planing to add new features to Byte Buddy for defining more complex constructors out of the box. 

Sentence removed

- [x] At the "Miscellaneous" section, I could not find these "Forwarding implementations". There is no class implementing the "Implementation" interface that is called "Forwarding".
- [x] Check commits that indicated methods such as defineField and defineMethod do not exist anymore, supposedly being replaced by define(Field) and define(Method). However, the "Accessing fields" section correctly shows such methods still exists. Maybe they are accessed from a different object. This needs to be reviewed.
- [x] There is no `TypeAttributeAppender.ForSuperType` and I couldn't find a replacement.
- [x] I couldn't find this `AnnotationFilter` type which is supposed to be given to a `ByteBuddy` instance.
- [ ] The lib has a large amount of types and sometimes it's difficult to know where a given type can be used. For instance, how can I use the `AnnotationRetention.DISABLED` to disable annotation retention? If some `@see` JavaDoc tags were included, it would make it easier to find out how to use a given feature.


manoelcampos:
The [link about inflation mechanism used for reflection](https://manoelcampos.github.io/byte-buddy/#/tutorial) is broken.
The only reference I could find was [this one](https://www.ibm.com/docs/en/ztpf/2023?topic=options-dsunreflectinflationthreshold).

However, this inflation mechanism seems to be dropped in [JDK 18](https://openjdk.org/jeps/416), being replaced by method handles (I'm not sure about that). This way, I think the sentence below needs to be rewritten:

> However, modern JVMs know a concept called inflation where the JNI-based method invocation is replaced by generated byte code that is injected into a dynamically created class.

manoelcampos:
# develop page

The page states:

> The project's implementation is found in the byte-buddy-dev module. This module is configured to rely on a direct dependency onto the ASM library, thus the dev qualifier in the module's name.

You probably meant `byte-buddy-dep` module.

raphw:
This is a great effort if you want to update the documentation. I did not find time to work on it in a long time, and it is rather outdated. It would really need some attention.

manoelcampos:
Hey @raphw 
I'm updating the tutorial page (the last one), but it will take some time. I hope I finish this week.
I updated the comments above and added some checkboxes to track the progress.
I'd need your help on those unchecked ones.

manoelcampos:
Hey @raphw,

I just need to finish reviewing the "Custom method implementations" on the tutorial page.
This will take some time. But since there are lots of changes already and some of them need your help, 
you can start reviewing when it's suitable for you.

Check an overview of the changes below:

- Performs comprehensive review of the entire documentation website
- Fix broken links
- Replaces reference to Travis CI by GitHub Action
 - Fix typos, grammar, punctuation and coherence issues.
 - Split long paragraphs and sections to make the reading smoother and clearer.
 - Fix issues on examples, mainly due to API changes.

manoelcampos:
That is it. The complete documentation website was revised and updated.
I’ll be glad to make any changes you request.

raphw:
Sorry for the delayed reply, this was just a lot to go through. And first of all: thanks for making that effort, this is something I wanted to do myself for quite some time, and I really appreciate it.

As for your questions.
1. Java is compiled to Java 5 and can be run on it. This because Java 5 is still around and Byte Buddy is often used for tooling. It does however contain a Java 8 version of all class files to make use of modern JVM features that do not support older class files formats. There is no CI run as Java 5 is not hosted on clouds, mainly for licensing issues.
2. All code is to be documented means that it is easier to validate that there is documentation. Without such possibility to fail the build, it would be easy to slip up.
3. Java 9 support is not experimental, this is an outdated statement when I first added support for the module system.
4. Byte Buddy does indeed still not handle type inference.
5. The benchmark 2a etc reference the table, but it seems like I removed the numbers. The first line is 1, second and third line are 2a and 2b, forth and fifth are 3a and 3b.
6. "Furthermore, Byte Buddy currently only takes loaded classes as its argument" is far from correct as it is now. Everything is abstracted as `TypeDescription` rather than `Class`. This is from early alpha versions.
7. The auxiliary type refers to the implementation of `Callable` that is not the generated class, but an auxiliary class that is also generated to implement the generated class.
8. You'd need to run: `MethodDelegation.withCustomConfiguration().filter(...)` for this.
9. Yes, Byte Buddy is considered feature complete when it comes to its scope, but it will of course follow the development of the JVM. The library does however not aim to do more, like more high level AOP.
10. Forwarding is long replaced by `MethodCall` which is much more feature rich.
11. It's long been `TypeAttributeAppender.ForInstrumentedType`.
12. That's been `AnnotationRetention` for a long time.
13. I fully agree that more *@see* annotations would be helpful.

manoelcampos:
> Sorry for the delayed reply, this was just a lot to go through. 

No worries. I'll need a few days to review your reply and make the appropriate changes.

manoelcampos:
I've made the required changes. Please let me know if there is anything else. 
There is just one item to confirm:

> 9. Yes, Byte Buddy is considered feature complete when it comes to its scope, but it will of course follow the development of the JVM. The library does however not aim to do more, like more high level AOP.

Just clarifying this one. The page says "Furthermore, we are planing to add new features to Byte Buddy for defining more complex constructors out of the box". 

So, should this sentence be removed? 

## Change log

> 1. Java is compiled to Java 5 and can be run on it. 

Fixed

> 3. Java 9 support is not experimental, this is an outdated statement when I first added support for the module system.

Sentence removed.

> 5. The benchmark 2a etc reference the table, but it seems like I removed the numbers. The first line is 1, second and third line are 2a and 2b, forth and fifth are 3a and 3b.

References included.

> 6. "Furthermore, Byte Buddy currently only takes loaded classes as its argument" is far from correct as it is now. Everything is 
abstracted as `TypeDescription` rather than `Class`. This is from early alpha versions.

Sentence removed.

> 7. The auxiliary type refers to the implementation of `Callable` that is not the generated class, but an auxiliary class that is also generated to implement the generated class.

Added "Byte Buddy creates a helper class that implements Callable" before "This helper class is called an AuxiliaryType".

> 8. You'd need to run: `MethodDelegation.withCustomConfiguration().filter(...)` for this.

In fact, it is `MethodDelegation.withDefaultConfiguration().filter(...)`. I changed the sentence to show that.

manoelcampos:
> All very good. And yes, the mentioned sentence can be removed.

Removed. That is it.

raphw:
Brilliant, thank you so much!

PR #1801, opened 2025-04-24
Improve ClassFileVersion javadoc

Marcono1234:
Tries to improve the Javadoc a bit by updating outdated or incorrect information.


byte-buddy-dep/src/main/java/net/bytebuddy/ClassFileVersion.java
```diff
@@ -268,8 +268,7 @@ public static ClassFileVersion ofJavaVersionString(String javaVersionString) {
     }
 
     /**
-     * Creates a class file version for a given major release of Java. Currently, all versions reaching from
-     * Java 1 to Java 9 are supported.
+     * Creates a class file version for a given major release of Java.
```
Marcono1234, line 271: Changed this because it looks like the restriction than only up to Java 9 is supported does not apply.

PR #1806, opened 2025-05-03
Added `PrimitiveComparison`

yukoba:
I added `PrimitiveComparison`. `PrimitiveComparison` compares two primitive values on the operand stack using `==`, `!=`, `<`, `>`, `<=`, or `>=`, and pushes the resulting boolean value onto the stack.


raphw:
Thanks for the addition. I retroactively realized that the frame generation does not work on this level, as it makes some assumptions. This is a weakness in the current generation concept. Fixing this would be a larger, API-breaking change which I cannot currently prioritize. For this reason, I had to roll back the PR, unfortunately. Frames can only be issued on the ByteCodeAppender level as of now, when it comes to a generic component that is unaware of its application.

PR #1816, opened 2025-05-21
Fix incorrect URL in effective POM for Maven modules

### LINKED ISSUE #718: The url in the pom (of the submodules) is invalid
Maven seems to append the module name to project's url thus it references a page that does not exist.

http://bytebuddy.net/byte-buddy
http://bytebuddy.net/byte-buddy-agent

Please set the correct url for the submodules as well or configure a server side redirect.

You can test this with `mvn site` and then checking the links on the dependency page.

Marcono1234:
Fixes #718

See https://maven.apache.org/ref/3.9.9/maven-model-builder/index.html#inheritance-assembly for a description of these attributes.

You can see the effect of this by running `mvn help:effective-pom` for the Maven modules and comparing the changes.


raphw:
Thanks, I did not even know that there was an option for this.

PR #1829, opened 2025-06-24
Use JVMS 4.2.2 "Unqualified name" definition to validate identifier

gdarmont:
Hey Rafael,

Following our [google group discussion](https://groups.google.com/g/byte-buddy/c/VXwBPt_-3AQ), I managed to take some time to improve identifier name validation.
I validated the implementation check against the current [OpenJDK parser code](https://github.com/openjdk/jdk21u-dev/blob/f55ff018b2f6a14e7f751d29f56205aba33e8b69/src/hotspot/share/classfile/classFileParser.cpp#L4664).

I created two additional methods. I hesitated to create a single method, to avoid the almost same loop in both methods. But I went for readability in this case. Let me know what you think.

Hope this helps,
Guillaume


raphw:
Thanks!

PR #1848, opened 2025-08-15
ParameterList - Cache the size and avoid iterator pattern

gsmet:
I have been working on a very large monolith using Hibernate ORM and these particular methods were quite high in the profiling I made.

Using this simple trick reduces things quite a lot, for what I think is an acceptable trade-off of the code looking a bit old school.

FWIW, we have been fighting against iterator() in Quarkus in critical hot paths as it generates more allocations than simply going through the list.

What is specific here is that size() is also quite slow in some cases, and it's also called by hasNext().

**This is before:**

Screenshot From 2025-08-15 15-46-38

Zoom on the interesting part:

Screenshot From 2025-08-15 15-46-53

**This is after:**

Screenshot From 2025-08-15 15-46-44

Zoom on the interesting part:

Screenshot From 2025-08-15 15-47-02

### DIFF
--- a/byte-buddy-dep/src/main/java/net/bytebuddy/description/method/ParameterList.java
+++ b/byte-buddy-dep/src/main/java/net/bytebuddy/description/method/ParameterList.java
@@ -79,7 +79,11 @@ public interface ParameterList<T extends ParameterDescription> extends Filterabl
          * {@inheritDoc}
          */
         public boolean hasExplicitMetaData() {
-            for (ParameterDescription parameterDescription : this) {
+            // cache the size and make sure to avoid iterators here
+            // this pattern reduces the number of allocations and also the CPU usage
+            int size = size();
+            for (int i = 0; i < size; i++) {
+                ParameterDescription parameterDescription = get(i);
                 if (!parameterDescription.isNamed() || !parameterDescription.hasModifiers()) {
                     return false;
                 }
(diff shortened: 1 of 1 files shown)


raphw:
Of course I remember!

I am curious: is it necessary to both avoid the iterator, and to avoid the call to "size"? I would expect iterators to normally be optimized by the JIT if they lie ion a critical code path. Generally speaking, I am happy to optimize here.

gsmet:
So, the `size()` makes it better all along. Avoiding the `iterator()` is just to avoid some additional allocations, given it becomes quite a hot path for us. Our experience in Quarkus is that it's good to avoid them in hot paths (but we still use them a lot elsewhere of course).

When we have a lot of entities, allocations are really in the way.

FYI, on my benchmarks with this generated application with 4000 entities, the Hibernate ORM enhancement + proxy generation with ByteBuddy is taking 40% of the CPU and allocations of the build.

This change only helps a bit and it was an easy one. For the rest, things are a lot more complex and I haven't yet figured out how to improve things in a meaningful way.

gsmet:
I'll have a look tomorrow!

gsmet:
@raphw from what I can see, things look fine after your patch. Thanks.

I had another small question: if I have a `TypePool` with some caching, are the types updated when I update the classes?

Basically, my question is that we are doing some operations in two steps and I would like to know if I can reuse the type pool from the first steps (I need to make sure the changes mades at the previous step are visible to the second step).

That would be my expectation but I want to make sure it's true.

raphw:
I refactored a little more as the initialization value of the field would be `0` and there is technically no guarantee that the non-final initializer value is set in a different thread, so this is more correct in the framework of the memory model, but it should work just fine.

As for the `TypePool`: No, there is no automatism in this, but you could update the `Cache` in a listener. The listener receives the `DynamicType` from which you can read a type description in the current state.

